PR #178, opened 2025-01-11
Avoid pushing Terms through arbitrary Python code

eb8680:
This refactoring PR changes the behavior of `evaluate`, `call` and `defterm` to be more conservative about when they attempt to push `Term`s through an arbitrary host-language callable. 

It also makes `defop` extensible via `singledispatch`.


eb8680:
> Let's add a test for the new behavior.

OK, I added one in `test_ops_syntax.py`

Issue #171, opened 2025-01-10
Binder annotations should be simplified, generalized and documented


eb8680:
Currently we use two special types `Bound` and `Scoped` to annotate `Operation` signature arguments with totally ordered relative scopes given by integers: 
```py
@defop
def Let(
  var: Annotated[Operation[[], S], Bound(0)],
  val: Annotated[S, Scoped(1)],
  body: Annotated[T, Scoped(0)]
) -> T:
  ...
```
This works well enough for many cases, but it's limited in expressivity and usability.

It should be possible to combine these into a single `Scoped` annotation and infer what is a variable and what is a value, use type variable symbols and `|` in place of integers to define a partial order on relative scopes, and annotate return types with `Scoped` to express the class of binding signatures from #122 :
```py
@defop
def Let(
  var: Annotated[Operation[[], S], Scoped[A]],
  val: Annotated[S, Scoped[B]],
  body: Annotated[T, Scoped[A | B]]
) -> Annotated[T, Scoped[B]]:
  ...
```
This annotation syntax has roughly the following rules:
- Scopes are represented by disjunctions of distinct type variables (e.g. `A` or `A | B`) and partially ordered by subset inclusion (e.g. `A <= A | B` and `A | B <= A | B | C` but not `A <= B` or `A | B <= B | C`)
- `Operation`-typed arguments (eg. `var` in `Let` above) tagged with `Scoped[<x>]` are captured in other arguments annotated with `Scoped[<y>]` where `x` and `y` are scopes such that `x <= y`. These can be replaced with fresh `Operation` symbols of the same type.
- When a captured variable with scope `x` appears free in a term with scope `y` such that `not x <= y` (e.g. if `var` appeared free in `val` in `Let` above), the instance in the term is distinct and is not captured by the binder.
- Captured variables in the same scope or in non-nested scopes must be distinct. Captured variables in nested scopes can be recursively replaced with fresh distinct variables.
- Captured variables that are included in the scope on the return annotation appear free in the result but are captured in its subterms
- Non-captured variables that appear free in a term whose scope includes a captured variable can be inferred to be in a scope disjunct with no captured variables (e.g. variables other than `var` appearing free in `body` of `Let` above are inferred to be part of `B` and will appear free in the result)

These rules are somewhat complicated to state, but implicitly specify a straightforward alpha-renaming procedure generalizing the current implementation of `_BaseOperation.__free_rule__` that succeeds if they can be satisfied and fails if they cannot.

[...]

PR #180, opened 2025-01-14
Move `_BaseOperation.__free_rule__` into specification

eb8680:
This refactoring PR removes `Operation.__free_rule__` and takes behavior that was previously part of `_BaseOperation.__free_rule__` and moves it into `defdata`.


eb8680:
I just can't get this to typecheck, no matter what I do...

eb8680:
OK, I think this is ready for review. #196 needs to be merged first so CI passes, though.

effectful/ops/syntax.py
```diff
@@ -273,7 +264,7 @@ def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
 
 
 @_CustomSingleDispatchCallable
-def defterm(dispatch, value: T) -> Expr[T]:
+def defterm(__dispatch: Callable[[type], Callable[[T], Expr[T]]], value: T):
```
jfeser, line 267: Make sure to update signature in `docs/source/effectful.rst`.

effectful/ops/syntax.py
```diff
@@ -298,21 +289,22 @@ def defterm(dispatch, value: T) -> Expr[T]:
...
 @_CustomSingleDispatchCallable
-def defdata(dispatch, expr: Term[T]) -> Expr[T]:
-    """Converts a term so that it is an instance of its inferred type.
+def defdata(
+    __dispatch: Callable[[type], Callable[..., Expr[T]]],
+    op: Operation[..., T],
+    *args,
+    **kwargs,
+) -> Expr[T]:
+    """Constructs a Term that is an instance of its semantic type.
```
jfeser, line 302: Ditto.

effectful/internals/base_impl.py
```diff
@@ -51,13 +32,20 @@ def __hash__(self):
...
+            return typing.cast(
+                Callable[Concatenate[Operation[Q, V], Q], Expr[V]], defdata
+            )(self, *args, **kwargs)
 
-    def __free_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> "Expr[V]":
-        from effectful.ops.syntax import Bound, Scoped, defdata, defop
+    def __fvs_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> tuple[
+        tuple[collections.abc.Set[Operation], ...],
+        dict[str, collections.abc.Set[Operation]],
+    ]:
```
jfeser, line 47: Let's add some documentation here. It's unclear to me exactly what this rule does now.

effectful/internals/base_impl.py
```diff
@@ -86,25 +74,38 @@ def __free_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> "Expr[V]":
...
-            min_scope = min(bound_vars.keys(), default=0)
-            scoped_args[min_scope] |= unscoped_args
-            max_scope = max(bound_vars.keys(), default=0)
-            assert all(s in bound_vars or s > max_scope for s in scoped_args.keys())
+        min_scope = min(bound_vars.keys(), default=0)
+        scoped_args[min_scope] |= unscoped_args
+        max_scope = max(bound_vars.keys(), default=0)
+        assert all(s in bound_vars or s > max_scope for s in scoped_args.keys())
 
         # recursively rename bound variables from innermost to outermost scope
```
jfeser, line 89: Why are we renaming bound variables? If this rule returns the free variables in each position in the arguments list, I would expect bound variables to be removed from these sets.

jfeser (changes requested):

effectful/ops/syntax.py
eb8680: Why are those handwritten signatures necessary? The code should be the sole source of truth.

effectful/internals/base_impl.py
eb8680: This comment is obsolete after the changes in this PR and should be removed.

effectful/internals/base_impl.py
eb8680: As of this PR, `Operation.__fvs_rule__` is supposed to return the variables appearing free in the scope of each positional and keyword argument that are newly bound by the `Operation`. The results are aggregated by `fvsof` to compute the free variables of a `Term`.  For example:
```py
@defop
def Let(
  var: Annotated[Operation[[], S], Bound(0)],
  val: Annotated[S, Scoped(1)],
  body: Annotated[T, Scoped(0)],
) -> T: ...

x, y = defop(int), defop(int)
val = y() + 1
body = x() + y() + 2
assert Let.__fvs_rule__(x, val, body) == ({x}, set(), {x})
assert fvsof(Let(x, val, body)) == {y}
```

effectful/ops/syntax.py
jfeser: Sphinx isn't good at detecting callable objects. The workaround seems to be to extend sphinx.

jfeser: See https://github.com/sphinx-doc/sphinx/issues/3783

effectful/internals/base_impl.py
jfeser: Should the first assert be `Let.__fvs_rule__(x, val, body) == (({x}, set(), {x}), {})`?

jfeser: I see. Is the idea that these are the variables that are safe to alpha-rename? I think including this description in the docstring is a good idea, since this is different from a standard free variables computation.

eb8680: In addressing #171 in a followup #208 to this PR, I've made further significant changes to binding. I think it makes more sense to update the documentation there, which I've done a first pass on.

effectful/internals/base_impl.py
eb8680: All of this is removed in #208

jfeser (approved):

Issue #161, opened 2025-01-08
Fresh operation names should be fresh by default


eb8680:
After #149, all `Operation`s created via `defop` are fresh (i.e. equality is object identity), but by default will reuse the same name strings:
```py
def op_sig(x) -> int: ...

op1 = defop(op_sig)
op2 = defop(op_sig, name="op")
assert op1 != defop(op_sig) and op1 != op2

# This is and should be true
assert op1.__name__ != op2.__name__

# This is but probably shouldn't be true
assert op1.__name__ == defop(op_sig).__name__  == defop(op).__name__
```
Names have no effect semantically, but it would be helpful for debugging purposes to generate fresh names for fresh operations.

PR #199, opened 2025-01-15
Give fresh names to Operations created from types or Operations

eb8680:
Addresses #161 

This PR makes `defop` create a fresh name string when it is passed an `Operation` or a `type`. As noted in #161, `Operation` names have no semantics; this is purely a usability change to make the results of printing a `Term` easier to interpret.

It also fixes a bug in `_BaseOperation.__eq__` that violates the freshness condition of `defop`.

### DIFF
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -3,6 +3,7 @@ import dataclasses
 import functools
 import types
 import typing
+import uuid
 from typing import (
     Annotated,
     Callable,
@@ -180,10 +181,13 @@ def _(t: Callable[P, T], *, name: Optional[str] = None) -> Operation[P, T]:
 
 @defop.register(Operation)
 def _(t: Operation[P, T], *, name: Optional[str] = None) -> Operation[P, T]:
+
+    @functools.wraps(t)
     def func(*args, **kwargs):
         raise NotImplementedError
 
-    functools.update_wrapper(func, t)
+    if name is None:
+        name = t.__name__ + "__SYM_" + uuid.uuid4().hex.split("-")[0]
     return defop(func, name=name)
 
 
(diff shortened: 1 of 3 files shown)


effectful/ops/syntax.py
```diff
@@ -180,10 +181,17 @@ def _(t: Callable[P, T], *, name: Optional[str] = None) -> Operation[P, T]:
...
 def _(t: Operation[P, T], *, name: Optional[str] = None) -> Operation[P, T]:
+
+    @functools.wraps(t)
     def func(*args, **kwargs):
         raise NotImplementedError
 
-    functools.update_wrapper(func, t)
+    if name is None:
+        name = (
+            getattr(t, "__name__", str(t))[:10000]
```
jfeser, line 191: Why not take the whole name? If length is a concern, 10000 chars is definitely already too long.

effectful/ops/syntax.py
```diff
@@ -180,10 +181,17 @@ def _(t: Callable[P, T], *, name: Optional[str] = None) -> Operation[P, T]:
...
+            + "__"
+            + uuid.uuid4().hex.split("-")[0]
```
jfeser, line 193: I don't think uuid does much for us here over using random.randint

jfeser (changes requested):

effectful/ops/syntax.py
eb8680: 10000 is an arbitrary upper bound to keep the names generated by `defop` from growing without limit

jfeser: Ah I see the concern. I think it would be better to keep the freshness id separate from the string name and regenerate the id when we copy an operation.

eb8680: I've found it useful in practice when debugging to have the name reflect the sequence of `defop`s that led to an `Operation`. I think it's fine as long as names have a bounded length and don't affect `Operation` equality semantics.

effectful/ops/syntax.py
eb8680: OK, I switched it to `randint`.

effectful/ops/syntax.py
jfeser: Ok, that's fine.

jfeser (approved):

PR #201, opened 2025-01-15
Compute sizesof(Distribution) without sampling

eb8680:
This PR adds a case for `torch.distributions.Distribution` atoms in `handlers.torch.sizesof` so that the free variables of a distribution can be computed without drawing a sample, which may be expensive in general. It also simplifies corresponding logic and removes some dead code in `handlers.indexed`.


jfeser (changes requested):
Needs lint fix

jfeser (approved):

PR #208, opened 2025-01-17
Simplify binding annotation syntax and semantics

eb8680:
Addresses #171 
Blocked by #180 

This PR replaces the current `Scoped`/`Bound` annotations with a single generalized `Scoped` annotation that uses sets of symbols ordered by subset inclusion instead of integers ordered by value, as sketched in #171.

It also moves much of the logic into `Scoped` itself, simplifying `_BaseOperation` significantly, even beyond the changes already included in #180.

### DIFF
--- a/docs/source/semi_ring.py
+++ b/docs/source/semi_ring.py
@@ -5,7 +5,7 @@ from typing import Annotated, ParamSpec, Tuple, TypeVar, Union, cast, overload
 
 import effectful.handlers.numbers  # noqa: F401
 from effectful.ops.semantics import coproduct, evaluate, fwd, handler
-from effectful.ops.syntax import Bound, Scoped, defop
+from effectful.ops.syntax import Bound, Scoped_, defop
 from effectful.ops.types import Interpretation, Operation, Term
 
 P = ParamSpec("P")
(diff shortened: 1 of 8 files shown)


docs/source/semi_ring.py
```diff
@@ -50,19 +52,19 @@ def __add__(self, other: "SemiRingDict[K, V]") -> "SemiRingDict[K, V]":
...
 
 @defop
 def Let(
-    e1: Annotated[T, Scoped(1)],
-    x: Annotated[Operation[[], T], Bound(0)],
-    e2: Annotated[S, Scoped(0)],
+    e1: Annotated[T, Scoped[A]],
+    x: Annotated[Operation[[], T], Scoped[B]],
+    e2: Annotated[S, Scoped[B]],
 ) -> S:
```
jfeser, line 68: Should this be `Annotated[S, Scoped[A]]`?

jfeser: Or maybe `e1` should not be annotated at all.

eb8680: Changed to `Annotated[S, Scoped[A]]`.

Issue #203, opened 2025-01-15
Distribution wrappers in `handlers.pyro` should be deprecated in favor of `defterm`


eb8680:
It should be possible to replace `PositionalDistribution` with `defterm` and `defdata` implementations for `TorchDistribution`s. The same things would also serve as a replacement for `NamedDistribution`, though we'd need an extra `Operation` there since the basic `Distribution` API doesn't support `__getitem__` or an equivalent.

eb8680:
#214 makes some progress toward this but doesn't yet fully resolve it.

I think the general solution would be a pair of operations `bind_dims` and `unbind_dims` in `handlers.torch` that subsume `handlers.torch.to_tensor` and `positional_distribution`/`named_distribution` from #214:
```py
T = TypeVar("T", bound=torch.Tensor | torch.distributions.Distribution | tree.Structure[torch.Tensor | torch.distributions.Distribution)

@defop
def bind_dims(
  value: Annotated[T, Scoped[A | B]], 
  *names: Annotated[Operation[[], int], Scoped[B]]
) -> Annotated[T, Scoped[A]]:
  ...

@defop
def unbind_dims(
  value: Annotated[T, Scoped[A]],
  *names: Annotated[Operation[[], int], Scoped[B]]
) -> Annotated[T, Scoped[A | B]]:
  ...
```
`bind_dims` is basically the same as our current `to_tensor` extended with logic for handling `call`. `unbind_dims` is a little trickier because it shouldn't distribute over operations that bind positional dimensions.

eb8680:
Removing from the milestone since #214 is sufficient for now and the general solution would be quite a bit more effort.

PR #212, opened 2025-01-27
Fix missing typing of to_tensor

jfeser:
`to_tensor` is public so should be type annotated.


eb8680 (approved):
I guess this is more correct than before, so it can be merged, but we should probably rename `to_tensor` to something more general (`bind_named`? `to_positional`? `materialize`?) and convert it into an `Operation` that binds its input variables, similar to `deffn`. Then we could use it as part of a solution to #203.

PR #214, opened 2025-01-27
Implement sizesof for named and positional distributions

### LINKED ISSUE #213: `sizesof` doesn't work on `NamedDistribution`s
```
x = defop(int, name="x")
d = NamedDistribution(dist.Normal(0, 1).expand([4]), [x])
assert sizesof(d) == {x: 4}
```
Got: `x == {}`

jfeser:
Closes #213


eb8680:
I think it would be better and not that much more involved to address the root cause #203 instead of special-casing `handlers.torch.sizesof`

jfeser:
Ok, I'll rework this into something that addresses #203

jfeser:
Handling `NamedDistribution` is tricky, because the named dimensions don't appear as arguments to `torch_getitem`. We would need to switch from wrapping the distribution to reconstructing it with named arguments. For most (all?) distributions, naming the batch dimensions of the parameters is the same as naming the batch dimensions of the sample results, so this approach should still work. It's certainly more involved than this PR, because we need to introspect the distributions and reconstruct them, which might need per-distribution logic.

effectful/handlers/torch.py
```diff
@@ -84,7 +84,8 @@ def _getitem_ellipsis_and_none(
     return torch.reshape(x, new_shape), new_key
 
 
-def sizesof(value: Expr) -> Mapping[Operation[[], int], int]:
+@functools.singledispatch
```
eb8680, line 87: This can be removed, right?

effectful/handlers/pyro.py
```diff
@@ -252,190 +265,264 @@ def __repr__(self):
...
+            return dist_constr(*named_args)
+        case _:
+            raise NotImplementedError
 
-        n_base = len(base_dist.batch_shape) + len(base_dist.event_shape)
-        self.naming = Naming.from_shape(self.indices.keys(), n_base)
 
-        super().__init__()
+@defop
+def positional_distribution(
```
eb8680, line 289: Hmm, why does `positional_distribution` needs to be a separate operation, as opposed to just using `to_tensor` on `defterm(d)`?

effectful/ops/syntax.py
```diff
@@ -280,7 +280,7 @@ def infer_annotations(cls, sig: inspect.Signature) -> inspect.Signature:
         # pre-conditions
         assert cls._check_has_single_scope(sig)
         assert cls._check_no_typevar_overlap(sig)
-        assert cls._check_no_boundvars_in_result(sig)
+        # assert cls._check_no_boundvars_in_result(sig)
```
eb8680, line 283: Instead of disabling this check for all operations, can we reenable it and make the signatures of the offending operations compliant? This check corresponds to a case our alpha-renaming code doesn't yet cover, so we shouldn't stop enforcing it in general.

effectful/handlers/pyro.py
```diff
@@ -252,190 +265,264 @@ def __repr__(self):
...
-    indices: Mapping[Operation[[], int], int]
-
-    def __init__(
-        self, base_dist: pyro.distributions.torch_distribution.TorchDistribution
-    ):
-        self.base_dist = base_dist
-        self.indices = sizesof(base_dist)
+@defop
+def named_distribution(
+    d: Annotated[TorchDistribution, Scoped[A]],
```
eb8680, line 270: Per my comment below on reenabling `Scoped._check_no_boundvars_in_result`, I think `named_distribution` is the only operation whose signature violates that condition, so changing the annotation on `d` here to
```py
d: Annotated[TorchDistribution, Scoped[A | B]]
```
should be sufficient to safely reenable the check without being an egregiously incorrect description of `named_distribution`.

effectful/handlers/pyro.py
```diff
@@ -252,190 +265,264 @@ def __repr__(self):
...
-        return self._to_positional(
-            self.base_dist.log_prob(self._from_positional(value))
-        )
+        return self._base_dist.log_prob(value)
 
     def enumerate_support(self, expand=True):
-        return self._to_positional(self.base_dist.enumerate_support(expand))
+        return self._base_dist.enumerate_support(expand)
 
+    def __getitem__(self, key: Collection[Operation[[], int]]):
```
eb8680, line 377: I think this should be removed, it's not part of the base `Distribution` API (though I wish it was) and we don't want specific `Term` representations to leak into user code by departing from their semantic types' APIs. People can just use `named_distribution` directly.

effectful/handlers/pyro.py
```diff
@@ -252,190 +265,264 @@ def __repr__(self):
...
-    def __init__(
-        self,
-        base_dist: pyro.distributions.torch_distribution.TorchDistribution,
-        names: Collection[Operation[[], int]],
-    ):
-        """
-        :param base_dist: A distribution with batch dimensions.
+@defterm.register(TorchDistribution)
+@defterm.register(TorchDistributionMixin)
+@functools.singledispatch
```
eb8680, line 383: Why the second layer of `singledispatch` here? Can we just use `defterm.register` and `defdata.register` directly? That way it's easier to add support for new distributions defined outside the library by adding `defterm` patterns in situ.

eb8680 (commented):
This mostly makes sense, I just have a few comments

effectful/ops/syntax.py
jfeser: Sorry, forgot about that. I think `named_distribution` doesn't pass this check, but I think its signature is correct.

effectful/handlers/pyro.py
jfeser: We could move this logic into `to_tensor`, sure. The default behavior of `to_tensor` doesn't work though, because vmap is restricted to only return tensors.

jfeser: As discussed, I'd like to leave this alone for now.

effectful/handlers/pyro.py
```diff
@@ -252,190 +264,280 @@ def __repr__(self):
...
-    """A distribution wrapper that lazily converts indexed dimensions to
-    positional.
+@defop
+def named_distribution(
+    d: Annotated[TorchDistribution, Scoped[A | B]],
+    *names: Annotated[Operation[[], int], Scoped[B]],
+) -> Annotated[TorchDistribution, Scoped[A | B]]:
+    match defterm(d):
+        case Term(op=_call, args=(dist_constr, *args)) if _call is call:
+            if dist_constr is dist.Independent:
```
eb8680, line 274: I don't think this special case handling of `Independent` is necessary. You just need to distribute `named_distribution` across the arguments:
```py
d = defterm(d) if not isinstance(d, Term) else d
if not (d.op is call and issubclass(d.args[0], TorchDistribution)):
  raise NotImplementedError

dist_constr, args = args[0], args[1:]

def _to_named(a):
  if isinstance(a, torch.Tensor):
    return Indexable(typing.cast(torch.Tensor, a))[tuple(n() for n in names)]
  elif isinstance(a, TorchDistribution):
    return named_distribution(a, *names)
  else:
    return a
 
return dist_constr(*tree.map_structure(_to_named, args))
```

effectful/handlers/pyro.py
```diff
@@ -252,190 +264,280 @@ def __repr__(self):
...
-    """
 
-    indices: Mapping[Operation[[], int], int]
+@defop
+def positional_distribution(
+    d: Annotated[TorchDistribution, Scoped[A]],
+) -> Tuple[TorchDistribution, Naming]:
+    match defterm(d):
+        case Term(op=_call, args=(dist_constr, *args)) if _call is call:
+            if dist_constr is dist.Independent:
```
eb8680, line 303: As in `named_distribution`, I don't think this special case handling of `Independent` is necessary. You just need to distribute `positional_distribution` across the arguments - something like this:
```py
d = defterm(d) if not isinstance(d, Term) else d
if not (d.op is call and issubclass(d.args[0], TorchDistribution)):
  raise NotImplementedError

dist_constr, args = args[0], args[1:]
naming = ...
indices = ...

def _to_positional(a):
  if isinstance(a, torch.Tensor):
    return to_tensor(a, indices)
  elif isinstance(a, TorchDistribution):
    return positional_distribution(a)[0]
  else:
    return a
 
return dist_constr(*tree.map_structure(_to_positional, args)), naming
```

eb8680 (approved):
Looks good, thanks for taking a pass at this! I think there are still a couple of easy simplifications of `positional_distribution` and `named_distribution` as noted in comments, but you can address them in a followup PR if this is blocking other work.

eb8680:
CI failure is just a randomized distribution test.

effectful/handlers/pyro.py
```diff
@@ -252,190 +264,280 @@ def __repr__(self):
...
+        case _:
+            raise NotImplementedError
+
 
-    def __init__(
-        self, base_dist: pyro.distributions.torch_distribution.TorchDistribution
-    ):
-        self.base_dist = base_dist
-        self.indices = sizesof(base_dist)
+class _DistributionTerm(Term[TorchDistribution], TorchDistribution):
```
eb8680, line 325: This should be registered with and accessed through `defdata`, but we can also address that in a followup PR.

Issue #226, opened 2025-02-02
source installation issue


qinqian:
Following the tutorial to install effortful from source code will raise error for dm-tree on Mac Pro:

```txt
  Building wheel for dm-tree (setup.py) ... error
  error: subprocess-exited-with-error

  × python setup.py bdist_wheel did not run successfully.
  │ exit code: 1
  ╰─> [68 lines of output]
      /opt/homebrew/Caskroom/miniconda/base/envs/effectful/lib/python3.12/site-packages/setuptools/_distutils/dist.py:270: UserWarning: Unknown distribution option: 'test_suite'
        warnings.warn(msg)
      running bdist_wheel
      running build
      running build_py
      creating build/lib.macosx-11.1-arm64-cpython-312/tree
      copying tree/sequence.py -> build/lib.macosx-11.1-arm64-cpython-312/tree
      copying tree/__init__.py -> build/lib.macosx-11.1-arm64-cpython-312/tree
      copying tree/tree_test.py -> build/lib.macosx-11.1-arm64-cpython-312/tree
      copying tree/tree_benchmark.py -> build/lib.macosx-11.1-arm64-cpython-312/tree
      running build_ext
      Traceback (most recent call last):
        File "/private/var/folders/2_/vr6ztx9d0yx6xy6kxndy39rc0000gn/T/pip-install-e5a3mwpa/dm-tree_b1ad5735be1c49c081033ae51e813395/setup.py", line 69, in _check_build_environment
          subprocess.check_call(['cmake', '--version'])
        File "/opt/homebrew/Caskroom/miniconda/base/envs/effectful/lib/python3.12/subprocess.py", line 408, in check_call
          retcode = call(*popenargs, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/opt/homebrew/Caskroom/miniconda/base/envs/effectful/lib/python3.12/subprocess.py", line 389, in call
          with Popen(*popenargs, **kwargs) as p:
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/opt/homebrew/Caskroom/miniconda/base/envs/effectful/lib/python3.12/subprocess.py", line 1026, in __init__
          self._execute_child(args, executable, preexec_fn, close_fds,
        File "/opt/homebrew/Caskroom/miniconda/base/envs/effectful/lib/python3.12/subprocess.py", line 1950, in _execute_child
          raise child_exception_type(errno_num, err_msg, err_filename)
[...]

PR #227, opened 2025-02-02
add dm-tree dep through conda 

qinqian:


eb8680:
Closed as out of scope per #226  and to avoid requiring `conda` for installation.

PR #230, opened 2025-02-04
Fix performance bug in vmap call

jfeser:
`vmap` calls `repr` on its function argument, which can be slow when we pass a term containing a large tensor. Avoid this cost with a wrapper that has a trivial `repr`.


jfeser:
With this change, our slowest test goes from ~12s to <1s.

eb8680 (approved):

PR #232, opened 2025-02-04
Clean up lint and test actions

jfeser:
- Update action versions
- Enable pip caching
- Stop running docs build with tests


eb8680 (commented):
If we're removing the docs build from the test workflow, we should add a separate CI build stage where a docs build happens in parallel alongside the others. Can you either add that here or open an issue to make sure we don't forget to do it later?

PR #247, opened 2025-03-04
Make `to_tensor` an operation

jfeser:
Allows partial evaluation to be part of a term.

### DIFF
--- a/effectful/handlers/pyro.py
+++ b/effectful/handlers/pyro.py
@@ -1,3 +1,4 @@
+import functools
 import typing
 from collections.abc import Collection, Mapping
 from typing import (
@@ -27,7 +28,7 @@ except ImportError:
 from typing_extensions import ParamSpec
 
 from effectful.handlers.torch import sizesof, to_tensor
-from effectful.ops.semantics import call
+from effectful.ops.semantics import typeof
 from effectful.ops.syntax import Scoped, defop, defterm
 from effectful.ops.types import Operation, Term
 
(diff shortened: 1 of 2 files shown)


jfeser:
Blocked by #246

effectful/handlers/torch.py
```diff
@@ -180,20 +170,70 @@ def reindex_flat_tensor(t):
...
+    result = tree.map_structure(reindex_flat_tensor, flat_result)
+    return result
+
+
+@defop
+def to_tensor_(
+    t: Annotated[torch.Tensor, Scoped[A | B]],
+    *args: Annotated[Operation[[], torch.Tensor], Scoped[A]],
+) -> Annotated[torch.Tensor, Scoped[B]]:
+    def _evaluate(expr):
```
eb8680, line 184: Why is there a duplicate of `evaluate` here? This seems to suggest we should be changing the interface of `_partial_eval` to be an `apply` rule. Also, should `_partial_eval` just be absorbed into the body of `to_tensor_`? When would it ever be correct to use `_partial_eval` on its own?

effectful/handlers/torch.py
```diff
@@ -180,20 +170,70 @@ def reindex_flat_tensor(t):
...
+        if not isinstance(o, Term) or o.op not in order_set
+    ]
+    dim_ops = [a.op if isinstance(a, Term) else None for a in dims]
+    perm = [dim_ops.index(o) for o in args] + reindex_dims
+    tensor = tensor.permute(perm)
+    return tensor[(slice(None),) * len(args) + tuple(dims[i] for i in reindex_dims)]
 
 
-def to_tensor(t: T, order: Collection[Operation[[], torch.Tensor]] | None = None) -> T:
+def to_tensor(
```
eb8680, line 226: Why do we need both `to_tensor` and `to_tensor_` instead of just a single `to_tensor` operation?

effectful/handlers/torch.py
```diff
@@ -180,20 +170,70 @@ def reindex_flat_tensor(t):
...
+) -> torch.Tensor:
     """Convert named dimensions to positional dimensions.
 
     :param t: A tensor.
     :type t: T
     :param order: A list of named dimensions to convert to positional dimensions.
                   These positional dimensions will appear at the beginning of the
                   shape.
-    :type order: Optional[Sequence[Operation[[], int]]]
+    :type order: Optional[Collection[Operation[[], int]]]
```
eb8680, line 236: nit: It seems like the type here in both the signature and the documentation should probably be `Sequence`, not `Collection` (which can be unordered).

effectful/handlers/torch.py
```diff
@@ -180,20 +170,70 @@ def reindex_flat_tensor(t):
...
+    # (or slices)
+    reindex_dims = [
+        i
+        for i, o in enumerate(dims)
+        if not isinstance(o, Term) or o.op not in order_set
+    ]
+    dim_ops = [a.op if isinstance(a, Term) else None for a in dims]
+    perm = [dim_ops.index(o) for o in args] + reindex_dims
+    tensor = tensor.permute(perm)
+    return tensor[(slice(None),) * len(args) + tuple(dims[i] for i in reindex_dims)]
```
eb8680, line 223: I'm not seeing the equivalent code in the previous version for this block at the end of `to_tensor_`. What is this logic doing that is not accomplished by `_partial_eval`'s `reindex_flat_tensor`? Why is the `permute` call above necessary? Is this meant to fix a bug?

effectful/handlers/torch.py
jfeser: I had a bad time trying to overload `apply` without breaking something else, so I gave up. 

`_partial_eval` has other callers, so I left it as its own function.

effectful/handlers/torch.py
jfeser: I didn't think that `Scoped` annotations worked inside collection types, so `to_tensor_` needed to have a different interface than `to_tensor`. I think merging them would be fine, but would change the API slightly.

eb8680: OK, I'm not seeing why we'd need the `to_tensor` interface at all in that case. The variadic `to_tensor_` signature seems nicer anyway since it's closer to `deffn`'s.

effectful/handlers/torch.py
jfeser: The distinction between the two functions is that `_partial_eval` is only concerned with the partial evaluation rule and `to_tensor` applies partial evaluation, then reorders dimensions. In the previous implementation, `_partial_eval` did both things and the permute was implicit in `ordered_sized_fvs`.

effectful/handlers/torch.py
eb8680: I only see one other caller (`_register_torch_op`). Should that be made to use `to_tensor_` instead, or should `_partial_eval` be absorbed into its body? Right now it seems like the logic is dispersed across a few different locations in a way that makes it hard to unwind.

effectful/handlers/torch.py
jfeser: Let's use the variadic interface then.

effectful/handlers/torch.py
eb8680: I looked at this a bit over the weekend, will put up a separate PR.

PR #252, opened 2025-03-13
Add a cache for term analysis rules

jfeser:
All terms now must implement the `apply_rule` method. `t.apply_rule(r)` is equivalent to
```
with handler({apply: lambda _, op, *args, **kwargs: r(t, *args, **kwargs)}):
    return evaluate(t)
```
but the results returned by `r` may be cached on the term and reused. `_TermRuleCache` implements such a cache.

In some of the rule implementations, a wrapper type is introduced to make it possible to distinguish between analysis results and ordinary values present in the term tree.


jfeser:
Ok. I'll think about other options, but I don't have a more principled design in mind right now.

PR #254, opened 2025-03-13
Fixes for distribution embedding

jfeser:
This a subset of #241 that contains only changes to distribution embedding.


effectful/handlers/pyro.py
```diff
@@ -330,29 +329,43 @@ def positional_distribution(
...
-    dist_constr, args = d.args[0], d.args[1:]
 
-    if not (
-        d.op is call
-        and (
-            issubclass(dist_constr, TorchDistribution)
-            or issubclass(dist_constr, dist.torch_distribution.TorchDistributionMixin)
-        )
-    ):
+    if not isinstance(d, _DistributionTerm):
```
eb8680, line 333: A check of this form should not appear in a semantic function - behavior should not depend on specific `Term` subtypes that are internal implementation details of `defdata`/`defterm`.

Is there a syntactic test we can apply instead, since the previous version is no longer valid? Alternatively, can we make the previous version valid again?

effectful/handlers/pyro.py
```diff
@@ -330,29 +329,43 @@ def positional_distribution(
...
+        *[_to_positional(a) for a in d.args],
+        **{k: _to_positional(v) for (k, v) in d.kwargs.items()},
+    )
+
+    assert new_d.event_shape == d.event_shape
+    return new_d, naming
+
+
+@functools.cache
+def _register_distribution_op(
```
eb8680, line 361: Why was this necessary? Was it something to do with type inference and `call`?

effectful/handlers/pyro.py
```diff
@@ -302,15 +303,8 @@ def named_distribution(
...
-    dist_constr, args = d.args[0], d.args[1:]
 
-    if not (
-        d.op is call
-        and (
-            issubclass(dist_constr, TorchDistribution)
-            or issubclass(dist_constr, dist.torch_distribution.TorchDistributionMixin)
-        )
-    ):
+    if not isinstance(d, _DistributionTerm):
```
eb8680, line 307: As mentioned in another comment, semantic functions should not depend on `Term` subtypes that are implementation details of `defterm`/`defdata`.

effectful/handlers/pyro.py
```diff
@@ -330,29 +329,43 @@ def positional_distribution(
...
         raise NotImplementedError
 
     indices = sizesof(d).keys()
     naming = Naming.from_shape(indices, len(shape))
 
     def _to_positional(a):
         if isinstance(a, torch.Tensor):
+            a_indices = sizesof(a)
+            assert len(a_indices) == 0 or set(a_indices.keys()) == set(indices)
```
eb8680, line 342: This check implies that all parameter tensors should be broadcasted up to the same named dimensions. Is that actually true? Would something like this fail?
```py
positional_distribution(Normal(locs[i], scales[j]))
```

effectful/handlers/pyro.py
jfeser: Yes, I think so, but I don't remember the precise problem. I also think the terms are simpler this way, and this change opens up the possibility of handling the distribution constructors.

effectful/handlers/pyro.py
jfeser: I think that this assert is incorrect, but the example here does actually work because Normal broadcasts the tensors internally. In general, I think we need to either ensure that parameter tensors are broadcasted up to the same named dimensions or we will need a separate renaming dict for each parameter. I think broadcasting the parameters is simpler.

Regardless, I'm going to add some tests for this case.

jfeser: #255 refactors the distribution tests to consider cases like this.

jfeser: #255 revealed some bugs, which should now be fixed by broadcasting all distribution parameters before converting them to positional.

eb8680 (approved):

PR #255, opened 2025-03-14
Add indexed distribution tests

jfeser:
Adds tests for distributions like `Normal(loc[i], scale[j])`, where the parameters are broadcastable but are not yet broadcasted. 

Refactors the testcase generation to support generating parameters with fully named batch dimensions and partially named batch dimensions.


eb8680 (approved):
Looks great, thanks!

PR #257, opened 2025-03-18
Introduce `bind_dims`/`unbind_dims`

### LINKED ISSUE #203: Distribution wrappers in `handlers.pyro` should be deprecated in favor of `defterm`

jfeser:
As discussed here https://github.com/BasisResearch/effectful/issues/203#issuecomment-2627794884, introduce `bind_dims`, `unbind_dims` as a unification of `to_tensor`, `positional_distribution` and `torch_getitem`, `named_distribution`.

Closes #203

### DIFF
--- a/docs/source/named_tensor_notation.ipynb
+++ b/docs/source/named_tensor_notation.ipynb
@@ -33,7 +33,7 @@
     "import torch\n",
     "from torch import tensor\n",
     "\n",
-    "from effectful.handlers.torch import sizesof, bind_dims\n",
+    "from effectful.handlers.torch import bind_dims, sizesof\n",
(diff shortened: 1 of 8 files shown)


eb8680 (approved):
Great to see the API shrink again!

jfeser:
I'm not a huge fan of the context resetting that I'm doing here, but it protects the potentially lazy distribution properties from any nonstandard interpretation shenanigans.

jfeser:
Also closes #263

PR #258, opened 2025-03-18
Register `_DistributionTerm` with `defdata`

### LINKED ISSUE #225: Refactor `_DistributionTerm` to use `defdata`
This should be registered with and accessed through `defdata`, but we can also address that in a followup PR.

_Originally posted by @eb8680 in https://github.com/BasisResearch/effectful/pull/214#discussion_r1937617939_

jfeser:
Closes #225

### DIFF
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -584,12 +584,6 @@ class _BaseOperation(Generic[Q, V], Operation[Q, V]):
         return tuple(result_sig.args), dict(result_sig.kwargs)
 
     def __type_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> type[V]:
-        sig = inspect.signature(self._default)
-        bound_sig = sig.bind(*args, **kwargs)
-        bound_sig.apply_defaults()
-
-        anno = sig.return_annotation
-
         def unwrap_annotation(typ):
             """Unwrap Annotated types."""
             return (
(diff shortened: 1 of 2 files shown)


effectful/handlers/pyro.py
```diff
@@ -534,8 +534,8 @@ def _embed_masked(d) -> Term[TorchDistribution]:
 @defterm.register(dist.LogisticNormal)
 @defterm.register(dist.Normal)
 @defterm.register(dist.StudentT)
-def _embed_loc_scale(d) -> Term[TorchDistribution]:
-    return _DistributionTerm(type(d), d.loc, d.scale)
+def _embed_loc_scale(d: TorchDistribution) -> Term[TorchDistribution]:
+    return defdata(_register_distribution_op(type(d)), d.loc, d.scale)
```
eb8680, line 538: Once the relevant `Term` constructor for a semantic type has been registered with `defdata`, you shouldn't even need to invoke `defdata` explicitly - all these calls to `defdata` in all of the `defterm` patterns here can be replaced with applications of the operation, with `defdata` invoked implicitly via `Operation.__free_rule__`:
```diff
- return defdata(_register_distribution_op(type(d)), d.loc, d.scale)
+ return _register_distribution_op(type(d))(d.loc, d.scale)
```

eb8680 (commented):
Great simplification - I think it can go even further

eb8680 (approved):

PR #260, opened 2025-03-28
Switch to pyproject.toml and uv

jfeser:


eb8680 (commented):
Seems like one of the test failures is a bug?
```
def test_defun_4():
        x = defop(int)
    
        with handler(eager_mixed):
    
>           @defterm

tests/test_handlers_numbers.py:190: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
effectful/ops/syntax.py:741: in __call__
    return self._func(self.dispatch, *args, **kwargs)
effectful/ops/syntax.py:772: in defterm
    return __dispatch(type(value))(value)
effectful/ops/syntax.py:957: in _
    body = value(
tests/test_handlers_numbers.py:195: in compose
    @defterm
effectful/ops/syntax.py:741: in __call__
    return self._func(self.dispatch, *args, **kwargs)
effectful/ops/syntax.py:772: in defterm
    return __dispatch(type(value))(value)
effectful/ops/syntax.py:957: in _
    body = value(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = _IntegralTerm(_BaseOperation(<function _.<locals>.func at 0x7fd1[278](https://github.com/BasisResearch/effectful/actions/runs/14134204195/job/39601887708?pr=260#step:6:279)be160>, name=int, freshening=[976950315]), (), {})

    @defterm
    def fg(x: int) -> int:
>       return f(g(x))
E       TypeError: '_BaseTerm' object is not callable

tests/test_handlers_numbers.py:197: TypeError
```

jfeser:
Hmm. This test passes locally, so it could be flaky.

eb8680:
I don't see why it would be nondeterministic? Let's see if it fails again after pulling `master`.

eb8680:
Same test still fails.

Issue #272, opened 2025-05-19
`evaluate` should convert non-`Term`s


jfeser:
`Term` objects can contain non-terms that are convertible to `Term`s. Ideally, evaluating such a term should do the appropriate term conversion so that any "hidden" operations can be handled. Another option is to change term construction so that these terms are always expanded.

jfeser:
Blocked by #271

PR #278, opened 2025-05-19
Move callable term conversion to a new function

### LINKED ISSUE #271: `defterm` of `Callable` values is too aggressive
`defterm` for callables is too aggressive. Most callables cannot be safely converted to terms or this isn't desirable. We should remove this case from defterm.

jfeser:
Closes #271

### DIFF
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -749,22 +749,6 @@ def defterm(__dispatch: Callable[[type], Callable[[T], Expr[T]]], value: T):
     :type value: T
     :returns: A term.
     :rtype: Expr[T]
-
-    **Example usage**:
-
-    :func:`defterm` can be passed a function, and it will convert that function
-    to a term by calling it with appropriately typed free variables:
-
-    >>> def incr(x: int) -> int:
-    ...     return x + 1
-    >>> term = defterm(incr)
-
-    >>> print(str(term))
-    deffn(add(int(), 1), int)
-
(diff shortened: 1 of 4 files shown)


eb8680 (approved):

effectful/ops/syntax.py
```diff
@@ -924,8 +908,25 @@ def __call__(self, *args: Expr, **kwargs: Expr) -> Expr[T]:
         return call(self, *args, **kwargs)  # type: ignore
 
 
-@defterm.register(collections.abc.Callable)
```
eb8680, line 927: Do we still need a `defterm` pattern for `Callable`? We don't want to trace the internals of a function by default, but we do want function application to go through `call` so we can add partial evaluation rules.

jfeser: No, I don't think so. That behavior comes from `CallableTerm`, which is introduced by `defdata`.

eb8680: Right, but doesn't addressing #272 require introducing `call` on all function applications? I was expecting something like this:
```py
@defterm.register(Callable)
def defterm_callable(v: Callable[P, T]) -> Callable[P, T]:
    return functools.partial(call, v)
```

jfeser: Does it? I think that would be new behavior. You should be able to get the effect of handling `call` by handling apply instead.

eb8680 (approved):

PR #293, opened 2025-06-27
Fix evaluation of operations that return reducible terms

### LINKED ISSUE #292: `evaluate` does not fully evaluate term
I would expect all of the following asserts to hold, but the last one does not. In the last case, `evaluate` returns the redex `x() + y()` rather than fully evaluating it.

```
import effectful.handlers.numbers
from effectful.ops.semantics import coproduct, evaluate, handler
from effectful.ops.syntax import deffn, defop

x, y, z = defop(int), defop(int), defop(int)
intp = {x: deffn(1), y: deffn(2), z: deffn(x() + y())}

with handler(intp):
    assert z() == 3

assert handler(intp)(z)() == 3

assert evaluate(evaluate(z(), intp=intp), intp=intp) == 3

assert evaluate(z(), intp=intp) == 3
```

jfeser:
Closes #292

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -47,7 +47,7 @@ def apply(intp: Interpretation, op: Operation, *args, **kwargs) -> Any:
 
     """
     if op in intp:
-        return intp[op](*args, **kwargs)
+        return handler(intp)(intp[op])(*args, **kwargs)
     elif apply in intp:
         return intp[apply](intp, op, *args, **kwargs)
     else:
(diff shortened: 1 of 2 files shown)


effectful/ops/semantics.py
```diff
@@ -47,7 +47,7 @@ def apply(intp: Interpretation, op: Operation, *args, **kwargs) -> Any:
 
     """
     if op in intp:
-        return intp[op](*args, **kwargs)
+        return handler(intp)(intp[op])(*args, **kwargs)
```
eb8680, line 50: I think this should probably use `effectful.internals.runtime.interpreter` rather than `handler`, which could in some cases lead to a `coproduct` of the active interpretation with itself.

eb8680 (approved):

Issue #296, opened 2025-06-27
Consider deprecating dependency on `tree`


eb8680:
We use the handy [`tree` utility library](https://tree.readthedocs.io/) in a number of places throughout the codebase. Unfortunately, `tree` is fundamentally incompatible with support for `Term`s that have semantic types that are subtypes of `tree.Structure` (i.e. `Mapping` or `Sequence` or nested versions thereof), as it cannot identify these as non-traversable.

We should be able to replace it by adding extra patterns for `defterm` and `defdata` and making `evaluate` and `apply` use them more aggressively.

PR #297, opened 2025-07-01
Correctly rename bound variables that appear as dict keys

jfeser:
`tree.map_structure` isn't correct for renaming bound variables because it does not map dict keys. Replaced with a new function `_map_structure_and_keys`.


eb8680 (approved):
I'll merge this since it fixes a bug, but I think the right thing to do is address #296 directly - `tree` just doesn't quite do what we want in many cases.

PR #300, opened 2025-07-08
Support polymorphic function and collection types in `defdata`

eb8680:
Blocked by #299 

This PR uses the type inference module added in #299 to enable the semantic embedding of `Term`s whose semantic type is a polymorphic function or collection that can be unified with an abstract type from `collections.abc`.

It replaces the existing implementation of `Operation.__type_rule__` with a simpler and more general version in terms of the `infer_return_type` function added in #299.

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -1,5 +1,6 @@
 import contextlib
 import functools
+import typing
 from collections.abc import Callable
 from typing import Any, TypeVar
 
@@ -294,7 +295,12 @@ def typeof(term: Expr[T]) -> type[T]:
     from effectful.internals.runtime import interpreter
 
     with interpreter({apply: lambda _, op, *a, **k: op.__type_rule__(*a, **k)}):
-        return evaluate(term) if isinstance(term, Term) else type(term)  # type: ignore
+        if isinstance(term, Term):
+            # If term is a Term, we evaluate it to get its type
+            tp = evaluate(term)
+            return typing.get_origin(tp) or tp  # type: ignore
+        else:
+            return type(term)
 
 
 def fvsof(term: Expr[S]) -> set[Operation]:
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -598,48 +598,13 @@ class _BaseOperation(Generic[Q, V], Operation[Q, V]):
         return tuple(result_sig.args), dict(result_sig.kwargs)
 
     def __type_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> type[V]:
-        def unwrap_annotation(typ):
-            """Unwrap Annotated types."""
-            return (
-                typing.get_args(typ)[0] if typing.get_origin(typ) is Annotated else typ
-            )
+        from effectful.internals.unification import infer_return_type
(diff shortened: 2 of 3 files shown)


effectful/ops/semantics.py
```diff
@@ -294,7 +296,24 @@ def typeof(term: Expr[T]) -> type[T]:
...
+                    if tp.__bound__
+                    else tp.__constraints__[0]
+                    if tp.__constraints__
+                    else object
+                )
+            if isinstance(tp, types.UnionType):
+                raise TypeError(
+                    f"Cannot determine type of {term} because it is a union type: {tp}"
+                )
+            return typing.get_origin(tp) or tp  # type: ignore
```
eb8680, line 314: These changes to `typeof` are necessary to preserve its current behavior under the change to `__type_rule__`. The call to `evaluate` above applies the new unification-based `__type_rule__` and may return a generic type, but we want `typeof` to continue to behave like Python's `type` which does not know anything about type variables or generic types.

effectful/ops/syntax.py
```diff
@@ -598,48 +598,28 @@ def __fvs_rule__(
...
+        return_anno = self.__signature__.return_annotation
+        if typing.get_origin(return_anno) is typing.Annotated:
+            return_anno = typing.get_args(return_anno)[0]
 
+        if return_anno is inspect.Parameter.empty:
             return typing.cast(type[V], object)
+        elif return_anno is None:
+            return type(None)  # type: ignore
+        elif not freetypevars(return_anno):
+            return return_anno
```
eb8680, line 617: This fast path should ensure that in cases where the result type can simply be read off the operation signature, which is typical, we incur no new performance overhead relative to current behavior.

effectful/ops/syntax.py
```diff
@@ -598,48 +598,28 @@ def __fvs_rule__(
...
-                    inspect.Parameter.VAR_KEYWORD,
-                ):
-                    arg = bound_sig.arguments[name]
-                    tp: type[V] = type(arg) if not isinstance(arg, type) else arg
-                    return drop_params(tp)
+        return_anno = self.__signature__.return_annotation
+        if typing.get_origin(return_anno) is typing.Annotated:
+            return_anno = typing.get_args(return_anno)[0]
 
+        if return_anno is inspect.Parameter.empty:
```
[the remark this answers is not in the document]
eb8680: These are the same, `inspect.Parameter.empty is inspect.Signature.empty`.

effectful/ops/syntax.py
```diff
@@ -1006,9 +989,7 @@ def _(op, *args, **kwargs):
 
     base_term = __dispatch(typing.cast(type[T], object))(op, *args_, **kwargs_)
     tp = typeof(base_term)
-    if tp is typing.Union:
-        raise ValueError("Terms that return Union types are not supported.")
-    assert isinstance(tp, type)
+    tp = typing.get_origin(tp) or tp
```
jfeser, line 992: I think this already happens in `typeof`.

jfeser (changes requested):
One minor change, otherwise looks good.

jfeser (approved):

PR #303, opened 2025-07-10
Bump minimum Python version to 3.12

### LINKED ISSUE #221: Consider requiring Python 3.12
Python 3.12 includes new generics syntax and other typing-related functionality that might be helpful in `effectful` and is not backward-compatible via `typing_extensions` in the usual way. We should consider requiring Python 3.12 or higher to take advantage of that, unless there are compelling reasons to retain support for Python 3.10/11.

This will also likely require switching away from Mypy (#220 )

eb8680:
Resolves #221


jfeser:
Good to merge once tests pass.

PR #304, opened 2025-07-11
Prevent XLA preallocation in CI builds

eb8680:
I'm seeing some bizarre segfaults in CI builds in #299 that I can't reproduce locally. This is an attempt to remove one possible cause by setting `XLA_PYTHON_CLIENT_PREALLOCATE=false` in the CI workflow, which should be a good idea in general, though I have no idea if it's going to help with #299 .


jfeser:
Why would we want this in general? It seems worse to run jax in a non-standard configuration.

PR #289, opened 2025-06-17
Add a `defterm` pattern for comprehension syntax overloading

eb8680:
Blocked by #288 #295 
Supercedes #242

This PR uses the new higher-order operation `effectful.ops.syntax.defstream` from #295 which has a straighforward correspondence to Python's `ast.GeneratorExp` along with the disassembler added in #288 to enable a robust, general `defterm` implementation for overloading and metaprogramming generator comprehensions.

Remaining tasks:
- [x] Unit tests for AST transformers
- [ ] Unit tests for new `defterm`

Issue #277, opened 2025-05-19
Add operation for `__getitem__` on tuples


jfeser:
We should have some support for tuple- (or perhaps sequence-, or `collections.abc....`) valued terms.

eb8680:
Related to #205

eb8680:
I think this is blocked by #296

Issue #251, opened 2025-03-13
Eliminate import side effects


jfeser:
The following program should work without needing a dummy import to run the side effects for `effectful.handlers.torch`.

```
import torch
from effectful.ops.syntax import defop

x = defop(torch.Tensor)

t = torch.randn((2, 3))[x()]
```

eb8680:
I think this is true for number types (resolved in #335), but the reason we defer the imports of `handlers.torch` and `handlers.jax` is that those two libraries themselves have import side effects that don't always play nicely with each other. In particular, XLA likes to allocate all GPU memory for itself immediately, even if you aren't actually doing a JAX computation.

jfeser:
I was hopeful that some kind of lazy import (https://peps.python.org/pep-0690/) could help, but unfortunately singledispatch functions require the whole module to be imported so they can access the class objects. I don't see a path towards fixing this.

PR #305, opened 2025-07-13
Add `defdata` patterns and `Operation`s for polymorphic collection and `Number` types

eb8680:
Addresses #277 #296 #251
~~Blocked by #299~~
Blocking #289 

This draft PR uses the type inference machinery in #299 #300 to add support in `effectful` for custom `Operation`s that take and return arbitrarily nested polymorphic collection types and higher-order functions. It does so mostly by adding `defdata` patterns and `Operation`s for the immutable collection APIs defined in the standard library [`collections.abc`](https://docs.python.org/3/library/collections.abc.html), and as a result is quite straightforward to extend incrementally as needed.

This PR is feature-complete and seems to work fine, but #299 needs to land on `master` and then this needs to be broken up into a sequence of self-contained smaller PRs. It also includes most of the work toward #296 as a prerequisite, and part of #251 (moving the meat of `handler.numbers` into `ops.syntax`) as a side quest.


eb8680:
Going to close this, the contents will land in separate PRs.

PR #320, opened 2025-08-04
Term Unification

jjcmoon:
Currently, writing transforms in Weighted involves quite some boilerplate for pattern matching/parsing. So solve this, I implemented a unification method on terms, making it considerably less painful to match on elaborate patterns. The unification method is rather straightforward, as we can use the infrastructure of Effectful to make patterns as open terms.

@jfeser We previously discussed the pros/cons of adding higher-level ops vs doing lower-level pattern matching. I believe this has the advantages of both approaches. For example, to support KL-divergence, we can method to help construct it:

def kl_divergence(x, p, q):
    log_p = log_prob(p, x)
    log_q = log_prob(q, x)
    return Sum({x.op: reals()}, jnp.exp(log_p) * (log_p - log_q))

This can be used by a user to avoid manually defining the divergence each time. But it can also by used to parse the divergence in a program transform. For example, if we are doing a transform of the divergence between two gaussians, we can write the following.

x, loc1, loc2, scale1, scale2 = (defop(Any) for _ in range(5))
with interpretation({}):  # to make sure pattern is not evaluated
    pattern = kl_divergence(x(), Normal(loc1(), scale1()), Normal(loc1(), scale2()))
match = mgu((pattern, term))
# do some program transform...

Note that this program transform works, even when the user manually defines a KL-divergence instead of using our kl_divergence method. Currently, the unification is still somewhat brittle as it e.g. cannot detect commutativity of addition/multiplication. Handling this would detect the patterns even better and would be an obvious next step.

### DIFF
--- a/effectful/internals/mgu.py
+++ b/effectful/internals/mgu.py
@@ -1,0 +1,50 @@
+from typing import Any
+
+from effectful.ops.syntax import deffn, syntactic_eq
+from effectful.ops.types import Term
+
+
+def mgu(*pairs, intp=None) -> dict | None:
+    if intp is None:
+        intp = {}
+    for t1, t2 in pairs:
+        intp = _mgu(t1, t2, intp)
+    return intp
+
+
+def _mgu(term1: Any, term2: Any, intp: dict) -> dict | None:
+    """
+    Unifier of term1 with term2.
+    (Unlike usual mgu, this does not unify bi-directionally.)
+    """
+    # Case 1: terms are identical -> pass
+    if syntactic_eq(term1, term2):
+        return intp
+
+    # Case 2: term1 is ground -> fail
+    if not isinstance(term1, Term):
+        return None
+
+    # Case 3: term1 is an open var -> assign term2 to it
+    if len(term1.args) == 0:
+        if term1.op in intp:
+            if syntactic_eq(term2, intp[term1.op]()):
+                return intp
+            else:
+                return None
+        else:
+            intp[term1.op] = deffn(term2)
+            return intp
+
+    # Case 4: term1 is structured term -> check inductively
+    if (
+        isinstance(term2, Term)
+        and term1.op == term2.op
+        and len(term1.args) == len(term2.args)
+    ):
+        for t1, t2 in zip(term1.args, term2.args, strict=False):
+            intp = _mgu(t1, t2, intp)  # type: ignore
+            if intp is None:
+                return None
+        return intp
+    return None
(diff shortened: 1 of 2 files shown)


eb8680:
Here's some code that uses `match` to do something similar. I agree this is not very pretty, but I think it could be cleaned up quite a bit with some changes to `Operation` and `Term`:
```py
def test_matching_reconstruct_kl():

    @defop
    def Normal(loc: float, scale: float, *, value: float) -> float:
        raise NotImplementedError

    @defop
    def exp(x: float) -> float:
        raise NotImplementedError

    @defop
    def integrate(x: Operation[[], float], f: Callable[[float], float]) -> float:
        raise NotImplementedError

    @defop
    def kl(x, logp: float, logq: float) -> float:
        return integrate(x, exp(logp) * (logp - logq))

    mul = (defop(float)() * 2.).op
    sub = (defop(float)() - 1.).op

    y = defop(float, name="y")
    match kl(
        y,
        Normal(0.0, 1.0, value=y()),
        Normal(1.0, 2.0, value=y()),
    ):
        case Term(integrate, (
                Operation() as x,
                Term(mul_, (
                    Term(exp_, (Term(p1, (loc1, scale1), {"value": value1}),)),
                    Term(sub_, (
                        Term(p2, (loc12, scale12), {"value": value12}),
                        Term(q, (loc2, scale2), {"value": value2})
                    )),
                )
            )
        )) if value1.op == value2.op == value12.op == x and \
                p1 == p2 == q == Normal and \
                mul_ == mul and sub_ == sub and exp_ == exp:
            assert scale1 > 0.0 and scale12 > 0.0 and scale2 > 0.0
        case _:
            assert False, "Expected kl operation in match" 
```

jjcmoon:
> This doesn't seem so bad to me? I guess it's subjective

Sure, this is subjective.

> > other words, the example I gave above, where a pattern is constructed by a function, cannot work. ... I feel a better option would be to do matching as a context manager. Something like
> 
> I guess I'm not sure we should want this to work, [...]

To me this was appealing because you can automatically derive the KLD parser from a function defining KLD, which gives a single source of truth and avoids bugs in the parsers. But if this is not the approach we want to take, I can close this PR and start doing the Weighted program transforms in the style of your KL example above.

jjcmoon:
Ok, are there changes to `Operation`/`Term` I could help with that would make the python `match` less verbose?

eb8680:
I think the first step would be to collect a few examples of `match` statements in your current work that either fail because of some interaction between the `Term` and `match` API or are particularly unpleasant-looking and open an `effectful` issue about them.

PR #323, opened 2025-08-14
Remove dependency of `ops` on `dm-tree` 

eb8680:
Addresses #296 

This PR removes all uses of `tree` from `effectful.ops`. It also folds the behavior of `ops.syntax._map_structure_and_keys` into `evaluate`.


effectful/ops/syntax.py
```diff
@@ -535,6 +533,10 @@ def __init__(
         self._freshening = freshening or []
         self.__signature__ = inspect.signature(default)
 
+    @property
+    def __isabstractmethod__(self) -> bool:
+        return False
```
jfeser, line 538: What is this for?

jfeser (approved):

PR #327, opened 2025-08-26
Introduce NotHandled replacement for NotImplementedError

### LINKED ISSUE #312: Remove special casing of `NotImplementedError` in operation definitions
It's not straightforward to raise `NotImplementedError` from an operation body, and this is sometimes necessary. For example, some numpyro distributions will raise it when accessing properties that are not supported.

We should switch to an effectful-specific exception e.g. `NotHandled`.

jfeser:
Closes #312

### DIFF
--- a/effectful/handlers/numbers.py
+++ b/effectful/handlers/numbers.py
@@ -147,16 +147,16 @@ class _RealTerm(_ComplexTerm, Term[numbers.Real]):
         raise ValueError("Cannot convert term to float")
 
     def __trunc__(self) -> numbers.Integral:
-        raise NotHandled
+        raise NotImplementedError
 
     def __floor__(self) -> numbers.Integral:
-        raise NotHandled
+        raise NotImplementedError
 
     def __ceil__(self) -> numbers.Integral:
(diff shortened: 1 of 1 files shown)


effectful/ops/types.py
```diff
@@ -29,28 +30,28 @@ class Operation[**Q, V](abc.ABC):
 
     @abc.abstractmethod
     def __eq__(self, other):
-        raise NotImplementedError
+        raise NotHandled
```
eb8680, line 33: I think the errors in `ops.types` should still be `NotImplementedError`s; these are not `Operation`s and `NotImplementedError` is the exception type expected from `abc.abstractmethod`s/`ABC`s.

effectful/ops/types.py
eb8680, line 1: All of the `raise NotHandled`s in `ops.types` should be reverted to `raise NotImplementedError`, these are not `Operation` calls.

eb8680 (changes requested):

eb8680 (approved):

PR #335, opened 2025-08-27
Remove `handlers.numbers`

eb8680:
Addresses #251 

This refactoring PR moves the contents of `handlers.numbers` into `ops.syntax`, and removes the `handlers.numbers` module entirely. This change is extracted from #305.


tests/test_handlers_numbers.py
eb8680, line 1: The contents of this test file were appended unmodified to `test_ops_syntax.py`.

jfeser (approved):

PR #340, opened 2025-08-27
Add jnp methods to _ArrayTerm

jjcmoon:
Adds the following methods to `_ArrayTerm`:

```
['all', 'any', 'argmax', 'argmin', 'argpartition', 'argsort', 'astype', 'choose', 'clip', 'compress', 'conj', 'conjugate', 'copy', 'cumprod', 'cumsum', 'diagonal', 'dot', 'max', 'mean', 'min', 'nonzero', 'prod', 'ptp', 'ravel', 'repeat', 'reshape', 'round', 'searchsorted', 'sort', 'squeeze', 'std', 'sum', 'swapaxes', 'take', 'trace', 'transpose', 'var']
```
So now we can do things on the Array terms such as

```Python3
i = defop(jax.Array)
jnp.sum(i(), axis=0) # works
i().sum(axis=0) # doesn't work currently
```

(If the monkey patch is not desirable, I can also manually add the above methods, but I figured this would be a lot of boilerplate.)

### DIFF
--- a/effectful/handlers/jax/_terms.py
+++ b/effectful/handlers/jax/_terms.py
@@ -224,11 +224,12 @@ class _ArrayTerm(Term[jax.Array]):
         raise TypeError("A free array is not iterable.")
 
 
-for name in jax.Array.__dict__:
+for name, func in jax.Array.__dict__.items():
     if (
         not name.startswith("_")
         and name in jnp.__dict__
         and name not in _ArrayTerm.__dict__
+        and hasattr(func, "__call__")
     ):
         setattr(_ArrayTerm, name, getattr(jnp, name))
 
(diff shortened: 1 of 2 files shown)


jfeser (approved):

eb8680:
I think it would be better to unroll that loop manually in the definition of `ArrayTerm`, and to add at least a smoke test for all the new attributes added. Just because the methods share a name with a function doesn't mean that they correspond one-to-one; the method may do some preprocessing of its arguments or have a different signature.

eb8680:
The problem is that there is no guarantee that a method is semantically equivalent to the function with the same name. The array class specification comes from NumPy, which has 20 years of backwards compatibility obligations standing between it and rational, concise API design. If you can find statements in the JAX documentation or codebase that guarantee that `an_array.<methodname>(*args, **kwargs)` is exactly equivalent to `<methodname>(an_array, *args, **kwargs)`, then by all means use that guarantee to eliminate the boilerplate. If you can't, then it shouldn't be assumed.

PR #341, opened 2025-08-27
Move docs requirements into pyproject toml

eb8680:
After #338 doc deployments are failing with [this error](https://github.com/BasisResearch/effectful/actions/runs/17277714474/job/49038619899):
```
  Extension error:
  Could not import extension sphinx_autodoc_typehints (exception: No module named 'sphinx_autodoc_typehints')
```
This should fix it by adding `sphinx_autodoc_typehints` to the docs dependencies and moving those to `pyproject.toml`.

### DIFF
--- a/docs/source/requirements.txt
+++ b/docs/source/requirements.txt
@@ -1,8 +1,0 @@
-sphinx
-sphinxcontrib-bibtex
-sphinx_rtd_theme
-myst_parser
-nbsphinx
-pyro-ppl
-torch
-dm-tree
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,12 +42,19 @@ jax = ["jax", "dm-tree"]
 numpyro = ["numpyro>=0.19", "dm-tree"]
 
 [dependency-groups]
-dev = [
+docs = [
     "sphinx",
     "sphinxcontrib-bibtex",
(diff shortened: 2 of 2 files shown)


jfeser:
We should add a docs build stage to CI, otherwise changes like these will be tough to test.

eb8680:
OK, I added a docs build to CI and fixed a few warnings

jfeser (approved):

.github/workflows/publish_docs.yml
```diff
@@ -23,4 +23,4 @@ jobs:
       uses: sphinx-notes/pages@v3
       with:
         documentation_path: ./docs/source/
-        requirements_path: ./docs/source/requirements.txt
+        pyproject_extras: docs
```
jfeser, line 26: Does this not require all extras to make all the modules importable?

jfeser (changes requested):

.github/workflows/publish_docs.yml
eb8680: The `docs` dependency group includes all extras, so it should work as expected, but the deployment workflow is opaque and only triggered on commits to `master` so I can't be sure.

jfeser (approved):

PR #342, opened 2025-08-28
Convert `syntactic_eq` to a singledispatch and add tensor support

### LINKED ISSUE #276: `syntactic_eq` is not compatible with the `==` operation on `jax.Array`s
Terms that contain literal jax arrays can't be compared for syntactic equality, because the `==` operation for jax arrays does not return a boolean.

jfeser:
Eliminates the need for a custom jax syntactic_eq in weighted.

Closes #276

### DIFF
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -9,7 +9,7 @@ import types
 import typing
 import warnings
 from collections.abc import Callable, Iterable, Mapping
-from typing import Annotated, Concatenate
+from typing import Annotated, Any, Concatenate
 
 from effectful.ops.types import Annotation, Expr, NotHandled, Operation, Term
 
@@ -1115,8 +1115,10 @@ iter_ = _IterableTerm.__iter__
 next_ = _IteratorTerm.__next__
 
 
-@functools.singledispatch
-def syntactic_eq(x, other) -> bool:
+@_CustomSingleDispatchCallable
+def syntactic_eq(
+    __dispatch: Callable[[type], Callable[[Any, Any], bool]], x, other
+) -> bool:
     """Syntactic equality, ignoring the interpretation of the terms.
 
     :param x: A term.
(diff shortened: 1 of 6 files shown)


effectful/ops/syntax.py
```diff
@@ -1164,6 +1140,39 @@ def syntactic_eq[T](x: Expr[T], other: Expr[T]) -> bool:
...
+@syntactic_eq.register
+def _(x: collections.abc.Mapping, other) -> bool:
+    return isinstance(other, collections.abc.Mapping) and all(
+        k in x and k in other and syntactic_eq(x[k], other[k])
+        for k in set(x) | set(other)
+    )
+
+
+@syntactic_eq.register
+def _(x: collections.abc.Sequence, other) -> bool:
```
eb8680, line 1168: Add a case for `str | bytes`, as these are considered `Sequence[str]`s.

effectful/ops/syntax.py
```diff
@@ -1164,6 +1140,39 @@ def syntactic_eq[T](x: Expr[T], other: Expr[T]) -> bool:
         return x == other
 
 
+@syntactic_eq.register
+def _(x: Term, other) -> bool:
+    if not isinstance(other, Term):
```
eb8680, line 1145: I think the case `isinstance(other, Term) and not isinstance(x, Term)` is now missing

jfeser: This is handled by the base case `x == other`.

effectful/ops/syntax.py
```diff
@@ -1161,7 +1139,49 @@ def syntactic_eq[T](x: Expr[T], other: Expr[T]) -> bool:
...
+        and all(syntactic_eq(a, b) for a, b in zip(x, other))
+    )
+
+
+@syntactic_eq.register(object)
+@syntactic_eq.register(str | bytes)
+def _(x: object, other) -> bool:
+    # if isinstance(other, Term):
+    #     return False
+    breakpoint()
```
eb8680, line 1183: Remove this `breakpoint()` and uncomment or remove the lines above

PR #346, opened 2025-09-02
Fix bug in `fvsof`

eb8680:
This PR fixes a small bug in `fvsof` that was inadvertently introduced in #343 and adds a regression test that would have caught it.

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -338,7 +338,8 @@ def fvsof[S](term: Expr[S]) -> collections.abc.Set[Operation]:
     def _update_fvs(op, *args, **kwargs):
         _fvs.add(op)
         bindings = op.__fvs_rule__(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


effectful/ops/semantics.py
```diff
@@ -338,7 +338,8 @@ def fvsof[S](term: Expr[S]) -> collections.abc.Set[Operation]:
     def _update_fvs(op, *args, **kwargs):
         _fvs.add(op)
         bindings = op.__fvs_rule__(*args, **kwargs)
-        for bound_var in set().union(*(bindings.args, *bindings.kwargs.values())):
+        for bound_var in set().union(*(*bindings.args, *bindings.kwargs.values())):
```
jfeser, line 341: `set(binding.args) | set(binding.kwargs.values())` is a bit simpler.

jfeser (approved):

Issue #348, opened 2025-09-08
Applying `defop` to method descriptor definitions should create instance-level `Operation`s


eb8680:
A recurring pattern of effect and handler definition that is not well-supported in `effectful` despite appearing in several downstream use cases is the need to specialize the behavior of multiple different copies of a single operation, where both the copies and their behaviors are distinguished from one another by some piece of data passed as an argument.

One potential solution to this is to define such operations as methods on classes that carry the data. Indeed, `ops.syntax.defop` can already be used in class bodies as a decorator for method descriptors, e.g.:
```py
class Foo[T]:
  @defop
  def bar(self, x: T) -> T:
    ...
```
However, the current behavior of `defop` here is to create a single class-level operation accessible via the class attribute `Foo.bar` and bind the instance argument `self` with `functools.partial` when the attribute is accessed on an instance `Foo().bar`:
```py
foo = Foo()
assert isinstance(Foo.bar, Operation)
assert not isinstance(foo.bar, Operation)
```
This allows a program invoking `bar` to use instances to refer to different behavior, but it still requires a single monolithic handler for `bar` that introspects the `self` argument to specialize.

A more interesting and arguably more Pythonic behavior for `defop` that better solves the problem described above would be for accessing the method descriptor `Foo.bar` on an instance of `Foo` to dynamically create a new instance-level `Operation` that is bound to the instance:
```py
foo1, foo2 = Foo(), Foo()
assert all(isinstance(bar, Operation) for bar in (Foo.bar, foo1.bar, foo2.bar))
assert foo1.bar == foo1.bar  # instance-level operations are created once per instance
assert Foo.bar != foo1.bar and Foo.bar != foo2.bar and foo1.bar != foo2.bar
```
The default behavior of an unhandled instance-level `Operation` would be to call the class-level `Operation`:
```py
with handler({foo.bar: Foo_bar_impl}):
  with handler({foo1.bar: foo1_bar_impl}):
    # foo1.bar is handled and does not behave like Foo.bar
    assert foo1.bar(x) == foo1_bar(foo1, x) != Foo.bar(foo1, x)
    # default behavior of unhandled foo2.bar is to call Foo.bar
    assert foo2.bar(x) == Foo.bar(foo2, x) != foo1.bar(x)
```

PR #351, opened 2025-09-12
Generate instance-level `Operation`s for bound methods

eb8680:
Addresses #348 

This PR upgrades the behavior of `Operation`s derived from methods, as outlined in #348. Specifically, when a method descriptor in a class body is decorated with `defop`, accessing the method for the first time on a new instance of the class will generate and return a fresh `Operation` that persists for the lifetime of the instance. Subsequent accesses will return the same `Operation`.

### DIFF
--- a/effectful/ops/types.py
+++ b/effectful/ops/types.py
@@ -510,14 +510,14 @@ class Operation[**Q, V]:
 def __apply__[**A, B](op: Operation[A, B], *args: A.args, **kwargs: A.kwargs) -> B:
     """Apply ``op`` to ``args``, ``kwargs`` in interpretation ``intp``.
 
-    Handling :func:`apply` changes the evaluation strategy of terms.
+    Handling :func:`Operation.__apply__` changes the evaluation strategy of terms.
 
     **Example usage**:
 
-    >>> @defop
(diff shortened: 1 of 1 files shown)


tests/test_ops_syntax.py
```diff
@@ -363,7 +359,8 @@ def test_defop_setattr_class() -> None:
     class MyClass:
         my_op: ClassVar[Operation]
 
-    @defop
+    @defop  # type: ignore
+    @staticmethod
```
jfeser: What happens if we fail to use `staticmethod`?

effectful/ops/types.py
```diff
@@ -422,60 +423,123 @@ def __fvs_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> inspect.BoundArgume
 
         return result_sig
 
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.__name__}, {self.__signature__})"
```
eb8680: Note I haven't changed `__repr__` or `__str__`, I just moved `__call__` down to be next to the modified `__init_subclass__` so the recursion there is easier to follow.

PR #355, opened 2025-10-06
Add `namedtuple` case in `evaluate`

### LINKED ISSUE #354: `evaluate` should work with `namedtuple`
Following #322 `evaluate` works with dataclasses but does not yet support `collections.namedtuple`s. Some of these may have odd edge case behavior, e.g. `jax.scipy.spatial.transform.Rotation` overrides `__len__`, which breaks the generic `namedtuple` interface.

eb8680:
Resolves #354 

This PR adds a pattern for `namedtuple` instance evaluation in `evaluate` that is general enough to cover `jax.scipy.spatial.transform.Rotation`. It includes a couple of unit tests that failed before this PR and pass after.


eb8680:
@cooijmanstim I added you as a reviewer. Let me know if there are other edge cases I should test.

cooijmanstim (approved):
LGTM and tests on the robotl side pass with this version. I left one suggestion for simplicity but it's up to you.

Issue #377, opened 2025-10-24
`Template.define` does not work with methods


kiranandcode:
It would be convenient to also support methods:

```python
class Agent:
    history = []
    task = 'Talk about space!'

    @Template.define
    def determine_action(self) -> str:
        """using {self.history} work out the next action to perform to achieve {self.task}"""

    def act(self):
        action = self.determine_action()
        self.history.append(action)


with handler(provider):
    agent = Agent()
    agent.act()

```

kiranandcode:
Pushed commit to #363 supporting this.

kiranandcode:
oh, yep, that makes sense.

eb8680:
Closing as resolved by #377

Issue #349, opened 2025-09-08
Subtyping `Operation` should enable more precise control of default behavior


eb8680:
Another poorly supported pattern of effect and handler definition that has recurred in several downstream use cases as well as in `effectful` itself is when large groups of`Operation`s share their behavior in some way.

For example, within `effectful` both `handlers.jax` and `handlers.torch` programmatically create many `Operation`s that share common behavior, but compacting this into a handler for `apply` would also capture `Operation`s outside this set and thus be too coarse-grained.

One potential solution to this is to allow the creation of subtypes of `Operation` that define an `apply`-like `Operation` specific to their instances:
```py
class TorchOperation[**P, T](Operation[P, T]):

  @defop
  @classmethod
  def apply(cls, op: "TorchOperation[P, T]", *args: P.args, **kwargs: P.kwargs) -> T:
    if op in get_interpretation():
      return get_interpretation()[op](*args, **kwargs)
    else:
      return cls.apply(op, *args, **kwargs)

  def __call__(self, *args: P.args, **kwargs: P.kwargs) -> T:
    return self.apply.__default_rule__(self, *args, **kwargs)
```

PR #378, opened 2025-10-24
Add support for methods in `Template.define` (#377)

kiranandcode:
Implements support for methods for `Template.define` closing #377 .


eb8680 (approved):
Making `Template` a subtype of `Operation` via #349 would be the ideal solution but given that #348 and #349 are proving somewhat complicated to resolve it makes sense to land this in the meantime.

Issue #380, opened 2025-10-25
`Template.define` - working with tools that output image


datvo06:
This tool doesn't work due to Pydantic's complain:
```python
@defop
def render_from_angles(azimuth: float, elevation: float, distance: float) -> list[Image.Image]:
    """Rendering the MuJoCo model from the given angles.
    Args:
        azimuth: The azimuth angle in degrees.
        elevation: The elevation angle in degrees.
        distance: The distance from the center of the model.
    Returns:
        A list of images.
    """
    mj_model, context = to_mujoco(
        self._build_top_model(self._load_module()), build_path=Path(self.out_dir) / "trajectory_build"
    )
    data = mujoco.MjData(mj_model)
    mujoco.mj_forward(mj_model, data)
    return render_trajectory(mj_model, [data], azimuth=azimuth, elevation=elevation, distance=distance)
```
Is it possible to make it work?

datvo06:
I got this:
```
File ~/Marc/RDL/robotl/ops/specs.py:204, in multimodal_template.<locals>.deco.<locals>.TemplateClass.__new__(cls, *args, **kwargs)
    202 def __new__(cls, *args, **kwargs):
    203     if _current_registry is not None:
--> 204         return _current_registry.get_or_synthesize(meta)(*args, **kwargs)
    205     else:
    206         raise RuntimeError(
    207             "Template must be used within multimodal_context(with a registry)"
    208         )
...
PydanticSchemaGenerationError: Unable to generate pydantic-core schema for <class 'PIL.Image.Image'>. Set `arbitrary_types_allowed=True` in the model_config to ignore this error or implement `__get_pydantic_core_schema__` on your type to fully support it.

If you got this error by calling handler(<some type>) within `__get_pydantic_core_schema__` then you likely need to call `handler.generate_schema(<some type>)` since we do not call `__get_pydantic_core_schema__` on `<some type>` otherwise to avoid infinite recursion.
```

PR #385, opened 2025-10-28
Add support for tools returning images

kiranandcode:
Addresses #380. Tools that return `Image.Image` or list[Image.Image] are handled specially and sent to the model, according to the API described here: https://platform.openai.com/docs/api-reference/responses/create#responses-create-input-input-item-list-item-function-tool-call-output-output

Minor nits are that `Tool.result_model` doesn't seem to be used anywhere in the code and the type signature is set to be a pydantic model. I've expanded the type to encompass a custom `ImageFunctionOutput` type (adapted from a type present in the OpenAI API, but not exported).

Example code:
```
@defop
def get_selfie(person: str) -> list[Image.Image]:
  "Find the selife image for the specified person."
  return [Image.open('/Users/kirancodes/Desktop/selfie.jpg')]


@Template.define(tools=[get_selfie])
def selfie_rating(name: str) -> str:
  "I'm {name}! Use provided tools to find my selfie and give me a nice message about how I look!"
  raise NotHandled

with handler(provider):
  print(selfie_rating('Kiran'))
```

Writing standalone tests for this functionality requires Tool.__call__ to be broken into smaller pieces as most of the changes are in `OpenAIAPIProvider`'s implementation.

### DIFF
--- a/effectful/handlers/llm/providers.py
+++ b/effectful/handlers/llm/providers.py
@@ -3,11 +3,11 @@ import dataclasses
 import inspect
 import io
 import string
-from collections.abc import Iterable, Mapping
-from typing import Any, Literal, get_type_hints
+import typing
+from collections.abc import Iterable, Mapping, Sequence
+from typing import Any, get_type_hints
 
 import pydantic
-from openai import BaseModel
 
 try:
     import openai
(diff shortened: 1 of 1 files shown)


jfeser (changes requested):
Can the conversion of tool results into json be performed by the pydantic model that we're building? Maybe this functionality can really just be a slightly smarter version of the model creation code. If we do that, we should pull the model creator out into an operation.

kiranandcode:
@jfeser updated the code to go through conversion via a pydantic model. Still not as clean as I'd like as I wanted to preserve the previous behaviour, which does some weird things like constructing a `{status: ..., result: ....}` json object to just stringify it.

jfeser (dismissed):
It's not important to preserve strange behavior. This looks fine to merge or you can clean it up further if you like.

Issue #388, opened 2025-11-03
Support asynchronous execution of `Template.__call__`


jfeser:
LLM API invocations are slow and many applications have significant parallelism between invocations that we should be able to exploit. 

If we choose to use Python's async support, we should avoid leaking async behavior into the rest of `effectful`. This means that the core functions in `effectful.ops` should not become async functions and we should not add async duplicates of these functions.

jfeser:
We now have three different approaches: async (#390), concurrent.futures (#400), and sample (#417).

eb8680:
As shown in a couple examples in #547 we can already use `asyncio` with `Template`s via `asyncio.to_thread`. I'm not sure there's much else to do here for the moment, so I'm closing it until we have more specific plans/goals.

PR #390, opened 2025-11-07
Alternative proposal for an async interface for effectful

kiranandcode:
On the ongoing discussion regarding async support for effectful (#388), in particular for LLM-based interfaces, this PR implements a minor modification that allows async coroutines to flow through the usual effectful boilerplate. To an end client, the effectful API would end up looking as follows:

```python
@defop
def add(x: int, y: int) -> int: raise NotHandled

@defop
def multiply(x: int, y: int) -> int: raise NotHandled

@defop
async def async_fetch_number(key: str) -> int: raise NotHandled

@defop
async def async_square(x: int) -> int: raise NotHandled

def client_sync_code():
    return multiply(add(1,2), 3)

async def client_code():
    a = await async_fetch_number('a')
    b = await async_fetch_number('b')
    doubled = multiply(add(a, b), 2)
    c = await async_fetch_number('c')
    c_squared = await async_square(c)
    result = add(doubled, c_squared)
    return result

async def fetch_handler(key: str) -> int: await asyncio.sleep(0.1); return int(ord(key))
async def square_handler(x: int) -> int: await asyncio.sleep(0.05); return x * x

interpretation = {
    async_fetch_number: fetch_handler, async_square: square_handler,
    add: (lambda x, y: x + y), multiply: (lambda x, y: x * y)
}


async def main():
    # without interpretation
    term = await client_code()
    term_sync = client_sync_code()
    print(f'term is {term}')
    print(f'term_sync is {term_sync}')
    # with interpretation
    with handler(interpretation):
        result = await client_code()
        result_sync = client_sync_code()
        print(f'result is {result}')
        print(f'result_sync is {result_sync}')

    # with interpretation
    with handler(interpretation):
        result = await evaluate(term)
        result_sync = evaluate(term_sync)
        print(f'evaluated result is {result}')
        print(f'evaluated result_sync is {result_sync}')
asyncio.run(main())
# term is add(multiply(add(async_fetch_number(a), async_fetch_number(b)), 2), async_square(async_fetch_number(c)))
# term_sync is multiply(add(1, 2), 3)
# result is 10191
# result_sync is 9
# evaluated result is 10191
# evaluated result_sync is 9
```

Pros:
- async code just works
- sync code is unchanged
- you can mix sync and async code without thinking about it

Cons:
- the deferred nature of async coroutines and how that interacts with the global runtime handler might be an easy source of footguns.

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -2,18 +2,13 @@ import collections.abc
 import contextlib
 import dataclasses
 import functools
+import inspect
 import types
 import typing
 from typing import Any
 
 from effectful.ops.syntax import defop
-from effectful.ops.types import (
-    Expr,
-    Interpretation,
-    NotHandled,  # noqa: F401
-    Operation,
-    Term,
-)
+from effectful.ops.types import Expr, Interpretation, Operation, Term
 
 
 @defop
@@ -243,7 +238,6 @@ def evaluate[T](expr: Expr[T], *, intp: Interpretation | None = None) -> Expr[T]
     6
 
     """
-    import inspect
 
     from effectful.internals.runtime import get_interpretation, interpreter
 


kiranandcode:
interaction with apply

effectful/ops/semantics.py
```diff
@@ -228,15 +243,53 @@ def evaluate[T](expr: Expr[T], *, intp: Interpretation | None = None) -> Expr[T]
...
     if isinstance(expr, Term):
-        args = tuple(evaluate(arg) for arg in expr.args)
-        kwargs = {k: evaluate(v) for k, v in expr.kwargs.items()}
-        return expr.op(*args, **kwargs)
+        # Check if we need async evaluation
+        args_evaluated = [evaluate(arg) for arg in expr.args]
+        kwargs_evaluated = {k: evaluate(v) for k, v in expr.kwargs.items()}
+
+        # Check if any evaluated args/kwargs are coroutines
+        has_coro = any(inspect.iscoroutine(a) for a in args_evaluated) or any(
```
eb8680: Can we push this logic into an `if isinstance(expr, Coroutine)` branch in `evaluate`:
```py
...
if isinstance(expr, Term):
    args = tuple(evaluate(arg) for arg in expr.args)
    kwargs = {k: evaluate(v) for k, v in expr.kwargs.items()}
    result = expr.op(*args, **kwargs)
    if isinstance(result, Coroutine):
        return await result
    else:
        return result
elif isinstance(expr, Coroutine):
    return await expr
...
```

effectful/ops/semantics.py
```diff
@@ -228,15 +243,53 @@ def evaluate[T](expr: Expr[T], *, intp: Interpretation | None = None) -> Expr[T]
     6
 
     """
+    import inspect
```
eb8680: nit: move import to module level, `inspect` is part of the standard library

effectful/ops/syntax.py
```diff
@@ -547,19 +549,46 @@ def __hash__(self):
...
+                        warnings.warn(
+                            "Operations should raise effectful.ops.types.NotHandled instead of NotImplementedError.",
+                            DeprecationWarning,
+                        )
+                        raise NotHandled
+                except NotHandled:
+                    # Return the term representation
+                    return typing.cast(
+                        Callable[Concatenate[Operation[Q, V], Q], Expr[V]], defdata
+                    )(await_, self, *args, **kwargs)
```
eb8680: It looks like this is the only place where `await_` is introduced, but I'm not sure if that's right? That may need to happen in `Operation.__call__` instead. We should clarify the invariants we're trying to maintain around `await_` and async `Operation`s.

effectful/internals/runtime.py
```diff
@@ -1,35 +1,34 @@
...
+import typing
 
 from effectful.ops.syntax import defop
 from effectful.ops.types import Interpretation, Operation
 
 
 @dataclasses.dataclass
 class Runtime[S, T]:
-    interpretation: "Interpretation[S, T]"
+    interpretation: "contextvars.ContextVar[Interpretation[S, T]]"
```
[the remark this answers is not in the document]
kiranandcode: oh, yep done!

kiranandcode:
- BaseAsyncOperation (CallableTerm)
- @defop __await__()

PR #404, opened 2025-11-20
draft PR to discuss MAKER impl in effectful-llm

kiranandcode:
Edit: updated with the latest changes to effectful, we can implement the logic in the following snippet:

```python
def predict_next_step(game_state: GameState) -> Step:
    ValidStep = build_validated_model(game_state)

    @Template.define
    def predict_next_step_inner(game_state) -> ValidStep:
        """
        Given the state of the game of towers of Hanoi as follows:
        {game_state}
        Predict the next step to complete the game (moving all disks to the rightmost tower).
        Give a reasoning for your prediction, and return the step following the format:
        <step>start,end</step>
        where start and end are zero-based indices for the towers to move. Be concise and avoid wordy answers.
        """
        raise NotHandled

    s = predict_next_step_inner(game_state)
    return (s.start, s.end)


def solve_hanoi(state: GameState):
    log = []

    for i in itertools.count():
        print(f"step {i} - {state}")
        with handler(KAheadSampler()), handler(RetryLLMHandler()):
            step = predict_next_step(state)
        # track the step at each point
        if new_state := state.apply(step):
            log.append((state, step))

        state = new_state or state
        state.visualise()
        if state.is_done():
            break
```

I don't think this necessarily should even live in effectful, but regardless it's an interesting case study to kick the tires on potential async/concurrent interfaces for effectful so making this draft PR to collect discussion about it.

relevant snippets from the implementation (and links to the paper):

## Algorithm 3
```python
class Agent:
    ...
    def has_no_red_flags(self, response: str) -> Step | None:
        if len(response) > 450.0:  return None
        step = self.parse_response(response)
        if not step: return None
        if not (0 <= step.start < len(self.game_state.towers) and 0 <= step.end < len(self.game_state.towers)):
            return None
        if step not in self.game_state.valid_steps():
            return None
        return step

    def get_vote(self):  # algorithm 3
        while True:
            resp = self.predict_next_step()
            if step := self.has_no_red_flags(resp):
                return step
```
### Algorithm 2
```python
class FirstToMoveAheadSelector:
    def do_voting(self) -> Step:  # algorithm 2
        # run n in parallel repeatedly until k come out in top
        while True:
            # submit a batch of votes
            for vote in futures.as_completed(
                Executor.submit(agent.get_vote) for agent in self.agents
            ):
                self.votes[vote] += 1
                max_other_votes = max(
                    (self.votes[o_vote] for o_vote in self.votes if o_vote != vote),
                    default=0,
                )
                if self.votes[vote] >= max_other_votes + self.k:
                    return vote
```
### Algorithm 1
```python
def solve_hanoi(state: GameState):
    log = []

    for i in itertools.count():
        print(f"step {i} - {state}")
        step = FirstToAheadMoveSelector(state).do_voting()
        # track the step at each point
        log.append((state, step))

        state = state.apply(step)
        state.visualise()
```

### DIFF
--- a/tests/test_maker.py
+++ b/tests/test_maker.py
@@ -1,0 +1,313 @@
+import itertools
+import logging
+import random
+import re
+from collections import Counter
+from typing import Optional
+
+from openai import OpenAI
+from pydantic.dataclasses import dataclass
+
+from effectful.handlers import futures
+from effectful.handlers.futures import Executor, ThreadPoolFuturesInterpretation
+from effectful.handlers.llm import Template
+from effectful.handlers.llm.providers import LLMLoggingHandler, OpenAIAPIProvider
+from effectful.ops.semantics import handler
+
+
+@dataclass
+class Step:
+    start: int
+    end: int
+
+
+@dataclass(frozen=True)
+class GameState:
+    """State of a game of towers of Hanoi where the initial state is a
+    set of towers. We use higher numbers to represesnt smaller
+    disks. So [1,2,3] is a valid tower. The towers are all stacked at
+    the left at the start (self.towers[0]), and the goal is to move
+    them to the rightmost tower (self.towers[-1])."""
+
+    size: int
+    towers: tuple[tuple[int, ...], ...]
+
+    @classmethod
+    def new(cls, size: int) -> "GameState":
+        towers = [[] for _ in range(size)]
+        towers[0] = list(reversed(range(size)))
+        towers = tuple(tuple(tower) for tower in towers)
+        return cls(size, towers)
+
+    def visualise_text(self):
+        max_disk = self.size
+        width = max_disk * 2 + 3
+        for i, tower in enumerate(self.towers):
+            print(f"\nTower {i}:")
+            for disk in reversed(tower):
+                disk_width = (disk + 1) * 2 - 1
+                padding = (max_disk - disk_width) // 2
+                print(" " * padding + "=" * disk_width + " " * padding)
+            print("=" * width)
+        print()
+
+    def visualise_image(self):
+        "Uses python graphics libraries to visualise the state of the hanoi game."
+        try:
+            from PIL import Image, ImageDraw
+        except ImportError:
+            return None
+        # Pillow-based visualization
+        tower_width = 150
+        disk_height = 30
+        base_height = 20
+        pole_width = 10
+        img_width = tower_width * len(self.towers)
+        img_height = disk_height * (self.size + 1) + base_height + 50
+
+        img = Image.new("RGB", (img_width, img_height), "white")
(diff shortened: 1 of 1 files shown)


tests/test_maker.py
```diff
@@ -0,0 +1,317 @@
...
+                    continue
+                if len(tower_i) == 0:
+                    continue
+                # if tower_i's disk is smaller than tower_j's topmost, then it is valid to move from tower i to j
+                if len(tower_j) == 0 or tower_i[-1] < tower_j[-1]:
+                    steps.append(Step(i, j))
+        return steps
+
+
+class MicroAgent:
```
eb8680: It seems like `MicroAgent` should just be a single stateless `Template` with a structured output type annotation?
```py
@Template.define
def predict_next_state(state: GameState) -> Step:
    """..."""
```
Here the behavior of `parse_response`, `has_no_red_flags` and `get_vote` should all be implicit in the annotation.

eb8680 (commented):
This is neat, it seems like there are several takeaways. Alongside this iterative solver, it would also be cool to implement a recursive solver that is allowed to call itself as a tool.

tests/test_maker.py
eb8680: `parse_response` should just be the default behavior of structured output generation. `has_no_red_flags` would follow either from a more specific static `Step` type or making it a Pydantic model with the runtime validation logic from `has_no_red_flags`.

tests/test_maker.py
```diff
@@ -0,0 +1,317 @@
...
+
+from effectful.handlers import futures
+from effectful.handlers.futures import Executor, ThreadPoolFuturesInterpretation
+from effectful.handlers.llm import Template
+from effectful.handlers.llm.providers import LLMLoggingHandler, OpenAIAPIProvider
+from effectful.ops.semantics import handler
+
+
+@dataclass(frozen=True)
+class Step:
```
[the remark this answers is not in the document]
kiranandcode: Done

tests/test_maker.py
```diff
@@ -0,0 +1,244 @@
...
+                    continue
+                if len(tower_i) == 0:
+                    continue
+                # if tower_i's disk is smaller than tower_j's topmost, then it is valid to move from tower i to j
+                if len(tower_j) == 0 or tower_i[-1] < tower_j[-1]:
+                    steps.append((i, j))
+        return steps
+
+
+class Step(ABC):
```
eb8680: This name collides in a confusing way with `type Step = tuple[int, int]` above.

jfeser:
Note that #479 should be reverted in this branch.

eb8680:
Closing in favor of version included in #694

PR #405, opened 2025-11-21
Fix `syntactic_eq` implementation for jax arrays

jfeser:
Previously didn't handle arrays of different shapes properly.

### DIFF
--- a/effectful/handlers/jax/_handlers.py
+++ b/effectful/handlers/jax/_handlers.py
@@ -281,7 +281,9 @@ def _indexed_func_wrapper[**P, S, T](
 
 
 @syntactic_eq.register
-def _(x: jax.typing.ArrayLike, other) -> bool:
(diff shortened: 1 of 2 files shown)


eb8680 (commented):
I'm confused about why this is necessary. Why do we want two arrays of incompatible shape to return `False` rather than raise a shape error, as it presumably did before?

jfeser:
I'd expect `syntactic_eq` to return `False` rather than raise whenever two terms differ. I'd only expect it to raise if the terms contain objects that don't implement `__eq__` and don't have an overload.

eb8680 (approved):
I can see that argument, but I guess I'd worry this could lead to shape errors persisting further along some code path than they did previously. I don't feel strongly enough to block this from landing on `master` so I'll go ahead and merge it.

Issue #395, opened 2025-11-13
LLM Module should support different providers.


datvo06:
Right now, we only support the OpenAI provider. Different projects might require using Claude (Anthropic), OpenRouter, etc:
```python
with AnthropicHandler():
      template_call()
```
One subtlety is that TokenTracker might depend on a specific provider, and whether we can detect incompatible TokenTracker/provider pairs.
@jfeser @kiranandcode What do you think?

jfeser:
Thanks for testing this!

Before we introduce a new collection of handlers, we should see if the chat completions API is sufficient for our needs. From a quick skim, it looks like it supports tool calling, multimodality, and structured output. I'd like to avoid adding more low-level handlers, in part because they're not easy to test automatically.

Would you be interested in trying to port the current `OpenAIAPIProvider`?

PR #411, opened 2025-11-24
Adds type-based encoding and support for legacy APIs

### LINKED ISSUE #409: Type-based encoding of LLM arguments
Template methods frequently need custom encoding logic for certain types - i.e rendering a `GameState` as an `ascii` grid rather than as a python `__str__`. Overriding `__str__` is not ideal as users may want to still print the state as normal, only changing how it is presented to the LLM.

The encoding mechanism inside the llm module should be type-directed (singledispatch) and an effect (for provider specific encodings).

This may subsume the domain specific Image encoding that is currently provided by effectful, and allow overriding the effect to handle provider-specific conventions

kiranandcode:
Closes #409 and works towards #395 .

Adds operation:

```python
@defop
@singledispatch
def format_value(value: Any) -> list[MessagePart]:
```

and operation:

```python
@defop
def format_message_parts_to_openai(
    parts: list[MessagePart],
) -> Any:
```

plus modifies `OpenAIProvider` with subclass:

```python
LegacyOpenAIAPIProvider
```

### DIFF
--- a/effectful/handlers/llm/providers.py
+++ b/effectful/handlers/llm/providers.py
@@ -73,6 +73,16 @@ def _(value: Image.Image) -> OpenAIMessageContent:
     return [_pil_image_to_openai_image_param(value)]
 
 
+@format_value.register(str)  # type: ignore
+def _(value: str) -> OpenAIMessageContent:
+    return [{"type": "text", "text": value}]
+
+
+@format_value.register(bytes)  # type: ignore
+def _(value: bytes) -> OpenAIMessageContent:
+    return [{"type": "text", "text": str(value)}]
+
+
 @format_value.register(Sequence)  # type: ignore
 def _(values: Sequence) -> OpenAIMessageContent:
     if all(isinstance(value, Image.Image) for value in values):


kiranandcode:
I also ran this latest PR using the llm integration tests on the other PR and so am confident it works

effectful/handlers/llm/providers.py
```diff
@@ -55,6 +55,32 @@ def _pil_image_to_openai_image_param(
...
+def _(value: Image.Image) -> OpenAIMessageContent:
+    return [_pil_image_to_openai_image_param(value)]
+
+
+@format_value.register(Sequence)  # type: ignore
+def _(values: Sequence) -> OpenAIMessageContent:
+    if all(isinstance(value, Image.Image) for value in values):
+        return [_pil_image_to_openai_image_param(value) for value in values]
+    else:
+        return [{"type": "text", "text": str(values)}]
```
jfeser: This will capture `str`, which isn't what we want.

jfeser (changes requested):

effectful/handlers/llm/providers.py
kiranandcode: What do you mean by capturing str? and what encoding would you suggest instead?

kiranandcode: done!

jfeser (approved):

PR #414, opened 2025-11-24
Allow `_BaseOperation` subclasses to have an overrideable `apply` method

### LINKED ISSUE #349: Subtyping `Operation` should enable more precise control of default behavior

jfeser:
Closes #349

### DIFF
--- a/effectful/ops/types.py
+++ b/effectful/ops/types.py
@@ -454,10 +454,6 @@ class Operation[**Q, V]:
     def __str__(self):
         return self.__name__
 
-    @property
-    def __isabstractmethod__(self) -> bool:
-        return getattr(self.__default__, "__isabstractmethod__", False)
-
     def __get__(self, instance, owner):
         if instance is not None:
             # This is an instance-level operation, so we need to bind the instance


effectful/ops/types.py
```diff
@@ -62,29 +380,129 @@ def __type_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> type[V]:
...
     def __repr__(self):
         return f"{self.__class__.__name__}({self.__name__}, {self.__signature__})"
 
+    def __str__(self):
+        return self.__name__
+
+    def __get__(self, instance, owner):
+        if instance is not None:
+            # This is an instance-level operation, so we need to bind the instance
+            return functools.partial(self, instance)
```
eb8680: Let's change this now to `types.MethodType(self, instance)`

effectful/ops/types.py
```diff
@@ -15,40 +24,349 @@ class NotHandled(Exception):
...
+        op = cls._define_callable(default, **kwargs)
+        op.dispatch = default._registry.dispatch
+        op.register = default._registry.register
+        return op
+
+    @define.register(_CustomSingleDispatchCallable)
+    @classmethod
+    def _defop_customsingledispatchcallable(cls, default, **kwargs):
+        sig = inspect.signature(functools.partial(default.func, None))
+        op = cls._define_callable(default, signature=sig)
```
eb8680: ditto: we should only be accessing `_define_callable` indirectly via type dispatch

effectful/ops/types.py
```diff
@@ -62,29 +380,129 @@ def __type_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> type[V]:
...
+    def __get__(self, instance, owner):
+        if instance is not None:
+            # This is an instance-level operation, so we need to bind the instance
+            return functools.partial(self, instance)
+        else:
+            # This is a static operation, so we return the operation itself
+            return self
+
+    @classmethod
+    def apply[**A, B](
```
eb8680: Maybe on the base `Operation`, `apply` should be a class property?
```py
class Operation:
  ...
  @property
  @classmethod
  def apply(cls) -> "Operation":
    from effectful.ops.semantics import apply
    return apply
```

effectful/ops/types.py
```diff
@@ -15,40 +22,330 @@ class NotHandled(Exception):
...
+
+        """
         raise NotImplementedError
 
-    @abc.abstractmethod
-    def __default_rule__(self, *args: Q.args, **kwargs: Q.kwargs) -> Expr[V]:
+    @define.register(classmethod)
+    @classmethod
+    def _define_classmethod(cls, default, **kwargs):
+        return Operation._ClassMethodOpDescriptor(cls.define, default.__func__)
```
eb8680: This is an error, so how is this still working? Is the `classmethod` path in `define` not being exercised in tests?

eb8680 (approved):
Good to merge once tests pass

PR #412, opened 2025-11-24
Implement `FirstToKAhead` Sampler

kiranandcode:
Closes #407 

```python
class MovieGenre(str, Enum):
    ACTION = "action"
    COMEDY = "comedy"
    DRAMA = "drama"
    HORROR = "horror"
    SCIFI = "sci-fi"
    ROMANCE = "romance"


@Template.define
def classify_genre(plot: str) -> MovieGenre:
    """Classify the movie genre based on this plot: {plot}
    Respond with exactly one of: action, comedy, drama, horror, sci-fi, romance"""
    raise NotImplementedError

plot = "A rogue cop must stop a terrorist group from detonating bombs across the city."
with handler(LLMLoggingHandler()), handler(OpenAIAPIProvider(openai.OpenAI())):
    with handler(KAheadSampler()):
        genre = classify_genre(plot)
    print(f"\nsampled genre: {genre}")
```

PR #415, opened 2025-11-25
Implemented `ReducedKAheadSampler`

kiranandcode:
This PR builds on top of #412. Often we would like to use an LLM to generate types that can not be hashed, or whose outputs do not lie within a small enough set such that voting is likely to complete in any reasonable amount of time (think generating anything with a string component, or callables). Yet these outputs may still have some "essense" in a smaller set which would still provide useful signal for voting.

This PR implements a `ReducedKAheadSampler` which allows the user to provide a `reducer` function which maps LLM generated outputs to a reduced key-set for voting, and then runs the usual k-ahead sampling algorithm.
```python
class MovieGenre(str, Enum):
    ACTION = "action"
    COMEDY = "comedy"
    DRAMA = "drama"
    HORROR = "horror"
    SCIFI = "sci-fi"
    ROMANCE = "romance"

@dataclass(frozen=True)
class MovieClassification:
    genre: MovieGenre
    explanation: str = Field(..., description="explanation for the given movie classification")

@Template.define
def classify_genre(plot: str) -> MovieClassification:
    """Classify the movie genre based on this plot: {plot}"""
    raise NotImplementedError


plot = "A rogue cop must stop a terrorist group from detonating bombs across the city."

with handler(LLMLoggingHandler()), handler(OpenAIAPIProvider(openai.OpenAI())):

    def reducer(classification: MovieClassification) -> MovieGenre:
        return classification.genre
    def select_best(results: list[MovieClassification]) -> MovieClassification:
        return max(results, key=lambda res: len(res.explanation))
    
    sampler = ReducedKAheadSampler(reducer, select_best=select_best)
    classification = handler(sampler)(classify_genre)(plot)
    print(classification)
```

### DIFF
--- a/effectful/handlers/llm/sampling.py
+++ b/effectful/handlers/llm/sampling.py
@@ -1,4 +1,5 @@
-from collections import Counter
+from collections import Counter, defaultdict
+from collections.abc import Callable
 from concurrent import futures
 from concurrent.futures.thread import ThreadPoolExecutor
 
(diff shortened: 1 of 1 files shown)


eb8680:
Closing as not planned in this form, can be revisited later from scratch per the comments above

PR #416, opened 2025-11-26
Add inheritable class for stateful templates

### LINKED ISSUE #408: Support chat-based LLM interactions
To document the discussion from Friday, it would be useful to have access to the system/user roles mechanism to support chat-like state for LLMs

```python
class StatefulAgent(effectful.handlers.llm.Interactive):
    """You are a helpful agent who helps the user with their queries"""
   
    @Template.define
    def respond_to_user(user_message: str) -> str:
          """Returns an answer to the user's query: {user_message}"""

```

i.e here the openai interface will see the system/user log of the previous interactions it had with the user, rather than respond_to_user being pure.

jfeser:
This PR conflicts with #411 because they both do some refactoring to providers. The operation breakdown in this PR should continue to be reasonable, but the implementation of the operations may not.

Adds a class `Agent` that when inherited from makes any `Template` methods stateful. This is more subtle than it looks because invoking a template may cause other template invocations via tool calls. This implementation uses an effect `Agent.current_agent` that it sets to determine when to trigger state injection.

Closes #408

### DIFF
--- a/effectful/handlers/llm/agent.py
+++ b/effectful/handlers/llm/agent.py
@@ -51,16 +51,14 @@ class Agent:
             setattr(cls, method_name, wrapper)
 
     def _format_model_input(self, template, other, *args, **kwargs):
-        with handler({Agent.current_agent: lambda: None}):
-            prompt = fwd()
+        prompt = fwd()
         if Agent.current_agent() is self:
             assert self is other
             prompt = self.state + prompt
         return prompt
 
     def _compute_response(self, *args, **kwargs):
-        with handler({Agent.current_agent: lambda: None}):
-            response = fwd()
+        response = fwd()
         if Agent.current_agent() is self:
             self.state += response.output
         return response


effectful/handlers/llm/providers.py
```diff
@@ -275,96 +275,122 @@ def _call_tool_with_json_args(
...
+
+
+class OpenAIAPIProvider(ObjectInterpretation):
+    """Implements templates using the OpenAI API."""
+
+    def __init__(self, client: openai.OpenAI, model_name: str = "gpt-4o"):
+        self._client = client
+        self._model_name = model_name
+
+    @implements(Template.__call__)
```
kiranandcode: This PR looks good, though it's surprising to me this class is OpenAIAPIProvider but there's barely any OpenAI specific details. This makes me worried that the OpenAI-specific dependencies  might be hidden elsewhere.

Or actually the the openai dependencies are in the default implementation of `format_model_prompt`, maybe we can move that into this class? as in keep the op, but move the impl into a handler provided by this interpretation.

kiranandcode (approved):

PR #417, opened 2025-11-26
Add a `sample` function that computes a batch of template results in parallel

jfeser:
The sampling methods in #415 and #412 could be mostly implemented on top of a parallel `sample` operation. You get a little less parallelism, since you would wait for all voters to complete before scheduling new ones.

### DIFF
--- a/effectful/handlers/llm/sampling.py
+++ b/effectful/handlers/llm/sampling.py
@@ -1,5 +1,4 @@
 from collections import Counter
-from collections.abc import Callable, Sequence
 from concurrent import futures
 from concurrent.futures.thread import ThreadPoolExecutor
 
--- a/effectful/handlers/llm/sampling.py
+++ b/effectful/handlers/llm/sampling.py
@@ -1,5 +1,7 @@
+import functools
 import threading
 from collections import Counter
+from collections.abc import Callable, Sequence
 from concurrent import futures
 from concurrent.futures.thread import ThreadPoolExecutor
 
@@ -7,7 +9,7 @@ from effectful.handlers.llm import Template
 from effectful.handlers.llm.providers import completion, tool_call
 from effectful.internals.runtime import get_interpretation, interpreter
 from effectful.ops.semantics import fwd, handler
-from effectful.ops.syntax import ObjectInterpretation, defop, implements
+from effectful.ops.syntax import ObjectInterpretation, implements
 
 
 class KAheadSampler[**P, T](ObjectInterpretation):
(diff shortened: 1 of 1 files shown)


jfeser:
Unfortunately, adding type annotations to `sample` seems to crash mypy. I can figure out a workaround if we decide this is worth merging.

jfeser:
I think it would be better to implement this slightly differently, so converting to draft for now.

effectful/handlers/llm/sampling.py
```diff
@@ -45,3 +47,55 @@ def n_votes_ahead():
                 tasks.append(executor.submit(interpreter(intp)(fwd), *args, **kwargs))
         executor.shutdown()
         return self.votes.most_common(1)[0][0]
+
+
+def sample(template, n):
```
kiranandcode: shouldn't sample have a type signature? or is this the issue from before with mypy?

effectful/handlers/llm/sampling.py
```diff
@@ -45,3 +47,55 @@ def n_votes_ahead():
...
+                return [t.result() for t in completed.done]
+
+    def _completion(*args, **kwargs):
+        lock.release()
+        result = fwd()
+        lock.acquire()
+        return result
+
+    def _tool_call(*args, **kwargs):
+        lock.acquire()
```
kiranandcode: what exactly are the guarantees that the `lock` here is intended to provide? is it that end users can think of `sample` as being equivalent to running template `n` times in sequence (in a non-deterministic order)?

effectful/handlers/llm/sampling.py
```diff
@@ -52,15 +54,48 @@ def sample(template, n):
     def in_nested_call() -> bool:
         return False
 
+    lock = threading.Lock()
```
kiranandcode: Given a dedicated commit for adding this lock, did you introduce this due to running into a threading issue in an example? It would be good to have this as a test case.

jfeser: Nope. The thought was to make `sample` safer by limiting parallelism to handlers to `completion`.

effectful/handlers/llm/sampling.py
jfeser: Ok, done.

effectful/handlers/llm/sampling.py
```diff
@@ -45,3 +47,55 @@ def n_votes_ahead():
...
+        result = fwd()
+        lock.acquire()
+        return result
+
+    def _tool_call(*args, **kwargs):
+        lock.acquire()
+        try:
+            result = fwd()
+        except Exception as e:
+            lock.release()
```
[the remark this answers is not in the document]
jfeser: > if fwd() doesn't raise an exception this won't release the lock no?
> 
> ```python
> with lock:
>     return fwd()
> ```

We actually don't want to. If the tool call completes successfully, we keep the lock up to the point that we start another `completion`. Releasing the lock here allows the remaining threads to proceed as this thread dies. The `fwd` in the `completion` handler can fail safely because we don't hold the lock.

This raises a good point though: what should we do when one of the threads fails? Right now we reraise the exception from the failed thread and you get no samples. Maybe instead you should get the successful samples?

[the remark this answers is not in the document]
jfeser: Agreed on tests.

jfeser:
If we decide to add a parallel `sample`, we should build on top of the parallelism in LiteLLM instead.

PR #420, opened 2025-12-02
Add LLM Integration tests to the workflows.

kiranandcode:
These tests check that the LLMProvider/Synthesis all work end to end on a small suite of simple programs.

### DIFF
--- a/tests/test_handlers_llm_provider.py
+++ b/tests/test_handlers_llm_provider.py
@@ -131,7 +131,7 @@ class TestLiteLLMProvider:
     """Tests for LiteLLMProvider basic functionality."""
 
     @requires_openai
-    @pytest.mark.parametrize("model_name", ["gpt-4o-mini", "gpt-3.5-turbo"])
+    @pytest.mark.parametrize("model_name", ["gpt-4o-mini", "gpt-5-nano"])
     def test_simple_prompt_multiple_models(self, model_name):
         """Test that LiteLLMProvider works with different model configurations."""
         with (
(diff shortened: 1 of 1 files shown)


kiranandcode:
Happy to take comments/iterate on the design a bit. 
- Am I making too many LLM calls? Too expensive models? 
-  We need to make a few gpt-4o calls to test out structured decoding as the cheaper models don't consistently work with that or tool calling. 
- Tried to keep to the simpler models where possible. Maybe some of these tests are redundant. 
- Maybe we could constrain the tests to only run when certain files are touched?

tests/test_handlers_llm_provider.py
```diff
@@ -0,0 +1,379 @@
...
+
+                    assert callable(count_func)
+                    # Test the generated function
+                    assert count_func("banana") == 3
+                    assert count_func("cherry") == 0
+                    assert count_func("aardvark") == 3
+            except SynthesisError as e:
+                if i < 2:
+                    continue
+                raise e
```
jfeser: Let's extract the retry logic into a decorator. We'll probably want something similar for other tests.

tests/test_handlers_llm_provider.py
```diff
@@ -0,0 +1,379 @@
...
+
+
+# Global state for tool calling tests
+evaluation_count = 0
+evaluation_results = []
```
jfeser: Tests shouldn't have global mutable state.

jfeser:
I doubt it. That can wait for another PR.

tests/test_handlers_llm_provider.py
kiranandcode: Done!

jfeser (changes requested):
The claude tests are skipped. Is that intentional?

kiranandcode:
oh yep they're skipped because I haven't added a key to the secrets yet.

PR #421, opened 2025-12-04
Check signature of returned Callable for ProgramSynthesisHander

### LINKED ISSUE #361: `effectful.handlers.llm.synthesis` should type-check generated code
Currently code is parsed and loaded, but not type checked. We should support type checking, at least optionally.

datvo06:
Closes #361

### DIFF
--- a/effectful/handlers/llm/synthesis.py
+++ b/effectful/handlers/llm/synthesis.py
@@ -100,12 +100,13 @@ class ProgramSynthesis(ObjectInterpretation):
         try:
             code_obj = compile(source_code, filename, "exec")
             exec(code_obj, gs)
-            if self.type_check:
-                self.verify_callable_signature(gs[last_decl.name], t)
-                # TODO: even more static analysis and type checking, adding type guards, etc.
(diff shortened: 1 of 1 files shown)


datvo06:
@jfeser  Thanks! I see, so we directly do the static analysis. Is the expected solution somewhere along this line?
```python
import tempfile
from mypy import api

def _type_check_code(self, code: str, expected_type: type) -> None:
    """Run mypy on generated code to verify type correctness."""
    # Write code to temp file for mypy
    with tempfile.NamedTemporaryFile(mode='w', suffix='.py', delete=False) as f:
        f.write(code)
        temp_path = f.name
    
    try:
        result = api.run([temp_path, '--no-error-summary', '--no-pretty'])
        stdout, stderr, exit_code = result
        
        if exit_code != 0:
            raise SynthesisError(f"Type check failed:\n{stdout}", code)
    finally:
        import os
        os.unlink(temp_path)
```

eb8680:
I think this may have been handled in #519

PR #422, opened 2025-12-04
Adding mypy check and test

### LINKED ISSUE #361: `effectful.handlers.llm.synthesis` should type-check generated code

datvo06:
A more detailed check and better type context construction for the Synthesis Handler.
Attempt to close #361 and subsumes #421

### DIFF
--- a/effectful/handlers/llm/synthesis.py
+++ b/effectful/handlers/llm/synthesis.py
@@ -299,11 +299,11 @@ def lexical_context_to_source(
         try:
             encoder = type_to_encodable_type(type(obj))
             if encoder.t is str:
-                source = encoder.encode(obj)
+                source = typing.cast(str, encoder.encode(obj))
                 sources.append(textwrap.dedent(source).strip())
             elif encoder.t is SynthesizedFunction:
                 if include_body:
-                    synth = encoder.encode(obj)
+                    synth = typing.cast(SynthesizedFunction, encoder.encode(obj))
                     sources.append(textwrap.dedent(synth.module_code).strip())
                 else:
                     try:
--- a/tests/test_handlers_llm.py
+++ b/tests/test_handlers_llm.py
@@ -1,3 +1,4 @@
+import typing
 from collections.abc import Callable
 from dataclasses import dataclass
 
@@ -68,7 +69,9 @@ class SingleResponseLLMProvider[T](ObjectInterpretation):
         # Decode using the encoding layer (like real LiteLLMProvider.decode_response)
         ret_type = template.__signature__.return_annotation
         encoder = type_to_encodable_type(ret_type)
-        decoded = encoder.decode(self.response, template=template)
+        decoded = encoder.decode(
+            typing.cast(typing.Any, self.response), template=template
+        )
         # Call type_check (like real LiteLLMProvider._call)
         return type_check(decoded, template)
 


effectful/handlers/llm/synthesis.py
```diff
@@ -19,13 +25,246 @@ def __init__(self, message, code=None):
         self.code = code
 
 
+def collect_type_sources(t: type, seen: set[type] | None = None) -> dict[type, str]:
```
jfeser: This might be useful context for the LLM, but it's going to be very hard to make this function return complete and correct code. As mentioned below, I think we can avoid using this function to produce context for the type checker.

effectful/handlers/llm/synthesis.py
```diff
@@ -19,13 +25,246 @@ def __init__(self, message, code=None):
...
+    # Add a type assertion to verify the function matches the expected type
+    # Extract function name from code
+    try:
+        module_ast = ast.parse(code)
+        last_decl = module_ast.body[-1]
+        if isinstance(last_decl, ast.FunctionDef):
+            func_name = last_decl.name
+            # Format the expected type for annotation (using imported names)
+            type_annotation = _format_type_for_annotation(expected_type)
+            source_parts.append(f"_check: {type_annotation} = {func_name}")
```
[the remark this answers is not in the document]
datvo06: I added the `SynthesizedFunction` to make sure the formatting is correct in the first place.

effectful/handlers/llm/synthesis.py
```diff
@@ -19,82 +26,425 @@ def __init__(self, message, code=None):
...
+            success, error_msg = run_mypy_check(code, referenced_types)
+            if not success:
+                raise SynthesisError(f"Type check failed:\n{error_msg}", code)
+
+        # Build globals dict by importing types from their original modules
         gs: dict = {}
+        for typ in referenced_types:
+            module = inspect.getmodule(typ)
+            if module is not None:
+                gs[typ.__name__] = typ
```
jfeser: Why do we do this? Presumably, the code already contains the imports we need because we built them for mypy. We should pull out the part of `run_mypy_check` that adds imports and do that unconditionally so we don't duplicate the behavior.

effectful/handlers/llm/synthesis.py
```diff
@@ -19,82 +26,425 @@ def __init__(self, message, code=None):
...
+
+        # Format parameter types for display
+        if param_types is None:
+            param_types_str = "*args, **kwargs"
+        elif not param_types:
+            param_types_str = "(no parameters)"
+        else:
+            param_types_str = ", ".join(
+                _format_type_for_annotation(t) for t in param_types
+            )
```
jfeser: Why not just `repr` the callable type?

tests/test_handlers_llm_provider_synthesis.py
```diff
@@ -0,0 +1,128 @@
...
+                    if i < n - 1:
+                        continue
+                    raise e
+
+        return wrapper
+
+    return decorator
+
+
+class LimitLLMCallsHandler(ObjectInterpretation):
```
kiranandcode: should we move these to a shared test library?

effectful/handlers/llm/synthesis.py
```diff
@@ -19,82 +28,408 @@ def __init__(self, message, code=None):
...
+    """Encodes LexicalContext to Python source code representation.
+
+    encode: Extracts source code/signatures for types, functions, and values.
+    decode: Executes source code to reconstruct a lexical context.
+    """
+
+    t = str
+
+    @classmethod
+    def encode(cls, context: LexicalContext) -> str:
```
eb8680: I would expect `EncodableLexicalContext.encode` to recursively call `type_to_encodable_type`, not implement custom serialization logic for a bunch of other types internally.

effectful/handlers/llm/synthesis.py
```diff
@@ -19,82 +28,408 @@ def __init__(self, message, code=None):
...
+        """Generate Python source code representations of items in the lexical context."""
+        from effectful.ops.types import Operation
+
+        sources: list[str] = []
+        seen_names: set[str] = set()
+
+        for name, obj in context.items():
```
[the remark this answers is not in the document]
datvo06: Just implemented and I can factor them out to separate PRs.

eb8680:
This was done in #536

PR #431, opened 2025-12-08
Add beam search example using thermometer continuations

jfeser:
Implements a beam search for programs that use `sample` and `factor` effects. 

A step toward implementing more sophisticated search strategies as in https://arxiv.org/pdf/2512.03571.

### DIFF
--- a/docs/source/beam.py
+++ b/docs/source/beam.py
@@ -89,9 +89,7 @@ class BeamCandidate[S, T]:
     def __lt__(self, other: "BeamCandidate[S, T]") -> bool:
         return self.score < other.score
 
-    def expand[**P](
-        self, model_fn: Callable[P, T], *args: P.args, **kwargs: P.kwargs
-    ) -> typing.Self:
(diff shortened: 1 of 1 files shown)


eb8680:
We could just land a version of this on `master` as a nice self-contained example, it's not connected to the LLM stuff per se.

docs/source/beam.py
```diff
@@ -0,0 +1,132 @@
...
+        ):
+            try:
+                result = model_fn()
+            except Suspend:
+                in_progress = True
+
+        return BeamCandidate(trace_intp.trace, factor_intp.score, in_progress, result)
+
+
+def beam_search[S, T](model_fn: Callable[[], T], beam_width=3) -> BeamCandidate[S, T]:
```
eb8680: Let's generalize this slightly to take a `Callable[P, T]`

docs/source/beam.py
eb8680: Add a docstring at the top of the file explaining what this example does and pointing to any relevant references including the paper from the description and the [thermometer continuations paper](https://dl.acm.org/doi/abs/10.1145/3236771).

PR #440, opened 2025-12-09
implemented type-based decoding

### LINKED ISSUE #432: Type-based decoding for Template results (generalises Constrained Decoding)
I believe we discussed this in person previously, but I couldn't find an issue about this. Currently, `effectful.llm` implements a type-driven encoding algorithm for mapping parameters to templates:

https://github.com/BasisResearch/effectful/blob/62e45a41f922c889988ecfee58c7bccc0827e2ac/effectful/handlers/llm/providers.py#L59-L61

This allows users to extend the encoding strategy for new types as well as implement handlers to allow provider-specific customisations - for example, encoding images in a specific way to conform to a particular provider's API.

We should provide an analogous interface for decoding results from the API - the current implementation is partially type directed, by virtue of `_pydantic_model_from_type`, but this only works for `pydantic` classes:
 
https://github.com/BasisResearch/effectful/blob/62e45a41f922c889988ecfee58c7bccc0827e2ac/effectful/handlers/llm/providers.py#L382

As we provide things like code generation, I think making this entire process type directed would allow for a more extensible and clean implementation. Program Synthesis could be a handler for a method:
```python
@defop
@decode_result.register(Callable)
def decode_callable(_):
   raise NotHandled
```

This would also allow supporting the synthesis of interfaces or `ABC`s as well. We could preserve the pydantic functionality by either making it the default, or by making it the decode operation for subclasses of `pydantic.BaseModel`, or `pydantic.dataclasses.dataclass`, or `Enum`.

kiranandcode:
Closes #432. 

This PR modifies type decoding to go through the operation `decode_result` which is a `_CustomSingleDispatchCallable`.

The default implementation calls a helper `decode_object[T](ret_type: type[T], content: str) -> T`.

There is an override for `Callable` such that by default attempting to decode a class with type `Callable` will raise NotHandled, overriding `decode_callable` allows people to implement program synthesis.

This PR thus also modifies `ProgramSynthesis` to override decode_callable.

### DIFF
--- a/effectful/handlers/llm/providers.py
+++ b/effectful/handlers/llm/providers.py
@@ -433,7 +433,7 @@ def _pydantic_model_from_object_type(typ: type) -> type:
 @pydantic_model_from_type.register(tuple)
 def _pydantic_model_from_named_tuple_type(typ: type) -> type:
     # check if NamedTuple
-    if issubclass(typ, tuple) and hasattr(typ, "_fields"):
+    if isinstance(typ, type) and issubclass(typ, tuple) and hasattr(typ, "_fields"):
         names = list(typ._fields)
         hints = get_type_hints(typ)
         types = [hints.get(n, object) for n in names]
@@ -499,7 +499,11 @@ def _decode_tuple[T](ret_type: type[T], result_str: str) -> T:
     tuple_elts = result.value.model_dump()  # type: ignore
 
     # if named tuple, instantiate the type again with the retrieved values
-    if issubclass(ret_type, tuple) and hasattr(ret_type, "_fields"):
+    if (
+        isinstance(ret_type, type)
+        and issubclass(ret_type, tuple)
+        and hasattr(ret_type, "_fields")
+    ):
         names = list(ret_type._fields)  # type: ignore
         hints = get_type_hints(ret_type)
         types = [hints.get(n, object) for n in names]


kiranandcode:
@eb8680 updated! I had to update the interface to have a `pydantic_model_from_type` function which maps python types to pydantic types that the provider API will accept. This is tied to a corresponding decoding function which then must recover the user's type from the encoded file

Support for tuples is recursive. It's not the most efficient as we json.dumps in recursive calls to match the interface that `decode_result` expects, but given that the expected nesting for tuples isn't expected to be too deep, this shouldn't be a bottleneck.

The following types are supported:
```python
class Color(str, Enum):
    """Color enum for testing."""

    RED = "red"
    GREEN = "green"
    BLUE = "blue"


@dataclass
class Person:
    """Person dataclass for testing."""

    name: str
    age: int
    height: float


class PointDict(TypedDict):
    """Point TypedDict for testing."""

    x: int
    y: int


class PointTuple(NamedTuple):
    """Point NamedTuple for testing."""

    x: int
    y: int


@Template.define
def generate_number_value() -> numbers.Number:
    """Generate a number value."""
    raise NotImplementedError


@Template.define
def generate_bool_value() -> bool:
    """Generate a boolean value."""
    raise NotImplementedError


@Template.define
def generate_color_enum() -> Color:
    """Generate a Color enum value."""
    raise NotImplementedError


@Template.define
def generate_person() -> Person:
    """Generate a Person dataclass."""
    raise NotImplementedError


@Template.define
def generate_point_dict() -> PointDict:
    """Generate a PointDict TypedDict."""
    raise NotImplementedError


@Template.define
def generate_plain_tuple() -> tuple[int, bool]:
    """Generate a pair of an int and a bool."""
    raise NotImplementedError


@Template.define
def generate_point_tuple() -> PointTuple:
    """Generate a PointTuple NamedTuple."""
    raise NotImplementedError


@Template.define
def generate_complex_tuple() -> tuple[PointDict, PointTuple]:
    """Generate a pair of an int and a bool."""
    raise NotImplementedError

```

I left out image as OpenAI doesn't support image generation via the chat.completions API. I wasn't able to easily find models that would consistently output Images in correct format

(I can somewhat coerce image generation by providing a pydantic model with appropriate descriptions, but this is more the LLM generating what it thinks a base64 encoded image will be rather than the provider actually generating images and encoding it properly)

PR #442, opened 2025-12-12
Implements a unified `encode`ing/`decode`ing pipeline for `llm`

kiranandcode:
This PR refactors the encoding/decoding pipeline to go through a unified `encode`/`decode` operation.

- `effectful/handlers/llm/encoding.py` implements an associated type morally equivalent to:
   ```lean
   structure Encodable (A: Type) where
       U: Type
       encode: A -> U
       decode: U -> A
   ```
   it provides a function `type_to_encodable_type` which does a sort of type-class resolution and returns an instance of `Encodable` for a given type `T`.

- `effectful/handlers/llm/provider.py` now uniformly constructs these `Encodable[T]` objects and uses their encode and decode operations when calling the LLM.

The high level pipeline to implement `Template.__call__(template, args)`:
- use `type_to_encodable_type` to obtain an `Encodable[T]` for each type in the signature of template
- use the associated encode operators to convert the arguments `args` to pydantic types
- use `serialize` to map pydantic types to `OpenAIMessage`s
- call provider with message
- while tool_calls:
   - use `decode` to map llm provided pydantic types back to python user types
   - call tool with arguments
   - use `encode` to map result to a pydantic type
   - use `serialize` to map result to OpenAIMessage
   - send back to LLM
 - take final call, use `type_to_encodable_type` to get `Encodable` for return type
 - use `decode` to map pydantic type to user type
 - return final result

I expect some discussion about the interface around `Tool` - for this initial version, I kept most of the API and logic the same, but we might want to restructure it a bit. 

Another aspect is the encoding type resolution might be a bit overengineered. Currently it descends into types such as `tuple[Image.Image, Image.Image]` to construct an `Encodable` that does the right thing for nested types as well.

### DIFF
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -24,7 +24,7 @@ def _pil_image_to_base64_data_uri(pil_image: Image.Image) -> str:
     return f"data:image/png;base64,{_pil_image_to_base64_data(pil_image)}"
 
 
-class _Encodable[T, U](ABC):
+class EncodableAs[T, U](ABC):
     t: type[U]
 
     def __init__(self, *args, **kwargs):
@@ -45,7 +45,7 @@ class _Encodable[T, U](ABC):
         return [{"type": "text", "text": str(value)}]
 
 
-class Encodable[T](_Encodable[T, type]):
+class Encodable[T](EncodableAs[T, type]):
     t = type
 
 
@@ -59,7 +59,7 @@ def type_to_encodable_type[T](
 
 @type_to_encodable_type.register(object)
 def _type_encodable_type_base[T](ty: type[T]) -> Encodable[T]:
-    class BaseEncodable(_Encodable[T, T]):
+    class BaseEncodable(EncodableAs[T, T]):
         t: type[T] = ty
 
         @classmethod
@@ -77,7 +77,7 @@ def _type_encodable_type_base[T](ty: type[T]) -> Encodable[T]:
 def _type_encodable_type_pydantic_base_model[T: pydantic.BaseModel](
     ty: type[T],
 ) -> Encodable[T]:
-    class EncodablePydanticBaseModel(_Encodable[T, T]):
+    class EncodablePydanticBaseModel(EncodableAs[T, T]):
         t: type[T] = ty
 
         @classmethod
@@ -96,7 +96,7 @@ def _type_encodable_type_pydantic_base_model[T: pydantic.BaseModel](
 
 
 @type_to_encodable_type.register(Image.Image)
-class EncodableImage(_Encodable[Image.Image, ChatCompletionImageUrlObject]):
+class EncodableImage(EncodableAs[Image.Image, ChatCompletionImageUrlObject]):
     t = ChatCompletionImageUrlObject
 
     @classmethod
@@ -142,7 +142,7 @@ def _type_encodable_type_tuple[T](ty: type[T]) -> Encodable[T]:
         tuple[*(enc.t for enc in element_encoders)],  # type: ignore
     )
 
-    class TupleEncodable(_Encodable[T, typing.Any]):
+    class TupleEncodable(EncodableAs[T, typing.Any]):
         t: type[typing.Any] = encoded_ty
 
         @classmethod
@@ -207,7 +207,7 @@ def _type_encodable_type_list[T](ty: type[T]) -> Encodable[T]:
         list[element_encoder.t],  # type: ignore
     )
 
-    class ListEncodable(_Encodable[T, typing.Any]):
+    class ListEncodable(EncodableAs[T, typing.Any]):
         t: type[typing.Any] = encoded_ty
 
         @classmethod


effectful/handlers/llm/providers.py
```diff
@@ -422,13 +455,18 @@ def decode_response[**P, T](template: Callable[P, T], response: ModelResponse) -
...
     ret_type = template.__signature__.return_annotation
-    if ret_type == str:
-        return result_str  # type: ignore[return-value]
+    encodable_ty = type_to_encodable_type(ret_type)
 
-    Result = _pydantic_model_from_type(ret_type)
-    result = Result.model_validate_json(result_str)
-    assert isinstance(result, Result)
-    return result.value
+    if encodable_ty.t == str:
```
eb8680: As mentioned above, it would be nice to encapsulate any special-casing for `str` within `type_to_encodable_type`, rather than having it appear in downstream code like this.

effectful/handlers/llm/providers.py
```diff
@@ -178,22 +226,23 @@ def push_current_text():
...
-                    assert not format_spec, (
-                        "image template parameters cannot have format specifiers"
-                    )
-                    push_current_text()
-                    prompt_parts.append(
-                        {
-                            "type": "image_url",
-                            "image_url": _pil_image_to_base64_data_uri(obj),
-                        }
+                # TODO(kg): should this use the serialize mechanism?
```
eb8680: Is this `TODO` a reference to your comment in Slack about formatting customization, or something else?

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+        image_url = image["url"]
+        if not image_url.startswith("data:image/"):
+            raise RuntimeError(
+                f"expected base64 encoded image as data uri, received {image_url}"
+            )
+        data = image_url.split(",")[1]
+        return Image.open(fp=io.BytesIO(base64.b64decode(data)))
+
+
+def _type_encodable_type_dataclass[T](ty: type[T]) -> Encodable[T]:
```
eb8680: This `dataclass`-to-`pydantic` logic looks OK at a glance, but dataclasses have a lot of edge cases and I would have hoped `pydantic` would do a lot of this for us? Is there not something like this available in `pydantic` already?

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+@type_to_encodable_type.register(numbers.Number)
+class EncodableNumber(_Encodable[numbers.Number, float]):
+    t = float
+
+    def __init__(self, _):
+        pass
+
+    @classmethod
+    def encode(cls, vl: numbers.Number) -> t:  # type: ignore
+        return float(vl)  # type: ignore
```
eb8680: I don't think this is right, not every `Number` can be invertibly coerced to a `float`. Maybe just remove the `numbers.Number` case for `type_to_encodable_type` for now, as long as base numeric types `int`, `float`, `bool`, and `complex` are all explicitly covered elsewhere.

eb8680 (commented):
This looks great. Two high-level questions in addition to the comments below:
1. I'm not seeing the `type_to_encodable_type` logic for the case when the user type `T` is a `pydantic.BaseModel` subtype, or any test cases. Is that case meant to be the `object` case where `encode` and `decode` are identity mappings? Can you add tests that exercise this logic?
2. What happened to `deserialize`? Have we decided to identify it with `pydantic.BaseModel.model_validate_json`?

effectful/handlers/llm/providers.py
```diff
@@ -58,33 +61,41 @@ def _pil_image_to_openai_image_param(
...
-    return [_pil_image_to_openai_image_param(value)]
+@serialize.register(dict)  # type: ignore
+def _(value: dict) -> OpenAIMessageContent:
+    if "url" in value and isinstance(value["url"], str):
+        return [
+            {
+                "type": "image_url",
+                "image_url": typing.cast(ChatCompletionImageUrlObject, value),
+            }
+        ]
```
jfeser: This treats any value that has a `url` key as an image?

effectful/handlers/llm/providers.py
kiranandcode: yes, it's w.r.t the slack comment.

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+def _type_encodable_type_base[T](ty: type[T]) -> Encodable[T]:
+    class BaseEncodable(_Encodable[T, T]):
+        t = ty
+
+        @classmethod
+        def encode(cls, vl):
+            return vl
+
+        @classmethod
+        def decode(cls, vl: t) -> T:  # type: ignore
```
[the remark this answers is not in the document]
kiranandcode: either Any, or I can try playing around with placing one of the generics here. I did iterate a few times on the design here to get even some type checks, as python and mypy doesn't easily support associated/existential types

effectful/handlers/llm/encoding.py
kiranandcode: so the default behaviour with `object` handles this. The main reason for this is to support:
```
@dataclass
class Person:
   profile_picture: Image.Image
   name: str
   age: int
```

Without this handler, submitting a `Person` dataclass would raise an exception as pydantic doesn't know how to handle these types. Given we provide support for images, for this initial version I implemented functionality to provide consistent behaviour even for nested images, but maybe this can be something we don't support.

effectful/handlers/llm/providers.py
```diff
@@ -94,32 +105,69 @@ def _(values: Sequence) -> OpenAIMessageContent:
...
-        ret_ty_origin = typing.get_origin(ret_ty) or ret_ty
+        encoded_ty = type_to_encodable_type(sig.return_annotation)
+        encoded_value = encoded_ty.encode(value)
+        return serialize.dispatch(encoded_ty.t)(encoded_value)  # type: ignore
 
-        return format_value.dispatch(ret_ty_origin)(value)  # type: ignore
-
-    @classmethod
-    def of_operation(cls, op: Operation[P, T], name: str):
+    @property
```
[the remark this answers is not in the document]
kiranandcode: will update!

effectful/handlers/llm/providers.py
```diff
@@ -94,32 +105,69 @@ def _(values: Sequence) -> OpenAIMessageContent:
...
+            result = tool_call(
+                template,
+                self.operation,
+                **params,
+            )
+            # serialize back to U using encoder for return type
+            encoded_ty = type_to_encodable_type(sig.return_annotation)
+            encoded_value = encoded_ty.encode(result)
+            # serialise back to Json
+            return serialize.dispatch(encoded_ty.t)(encoded_value)  # type: ignore
```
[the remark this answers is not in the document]
kiranandcode: I think `serialize(encoded_value)` would work, I figured as in this case we have the additional information about the types here, we could exploit it rather than just trying to recover it from the runtime values.

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+    __dispatch: Callable[[type[T]], Callable[..., Encodable[T]]], ty: type[T]
+) -> Encodable[T]:
+    origin_ty = typing.get_origin(ty) or ty
+    return __dispatch(origin_ty)(ty)
+
+
+@type_to_encodable_type.register(object)
+def _type_encodable_type_object[T](ty: type[T]) -> Encodable[T]:
+    # Check if it's a dataclass and redirect to dataclass handler
+    if dataclasses.is_dataclass(ty):
```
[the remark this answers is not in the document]
kiranandcode: oh, will do.

kiranandcode:
@eb8680 
>   1. I'm not seeing the `type_to_encodable_type` logic for the case when the user type `T` is a `pydantic.BaseModel` subtype, or any test cases. Is that case meant to be the `object` case where `encode` and `decode` are identity mappings? Can you add tests that exercise this logic?

Yes, will do.

>   2. What happened to `deserialize`? Have we decided to identify it with `pydantic.BaseModel.model_validate_json`?

Yes, from implementing this, a slight refinement on the design. Pydantic gives us:

serialize: U -> json
deserialize: json -> U

it could be that the serialize defop I've defined is misnamed, what it provides is actually serialize: U -> OpenAIMessage, for which we don't need an inverse.

effectful/handlers/llm/encoding.py
kiranandcode: yes, Image specifically - we currently transparently convert `PIL.Image.Image` into `ChatCompletionImageUrlObject` for sending to the OpenAI interface.

eb8680: Is that behavior supported by any of the providers, though? Like does the OpenAI API support structured outputs that include images? If not I'd rather remove this case for now.

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+            raise RuntimeError(
+                f"expected base64 encoded image as data uri, received {image_url}"
+            )
+        data = image_url.split(",")[1]
+        return Image.open(fp=io.BytesIO(base64.b64decode(data)))
+
+
+def _type_encodable_type_dataclass[T](ty: type[T]) -> Encodable[T]:
+    """Handle dataclass encoding/decoding with recursive field encoding."""
+    if not dataclasses.is_dataclass(ty):
```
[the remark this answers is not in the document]
kiranandcode: done

effectful/handlers/llm/providers.py
```diff
@@ -94,32 +105,69 @@ def _(values: Sequence) -> OpenAIMessageContent:
...
+        return parameter_model
 
+    def call_with_json_args(
+        self, template: Template, json_str: str
+    ) -> OpenAIMessageContent:
+        """Implements a roundtrip call to a python function. Input is a json string representing an LLM tool call request parameters. The output is the serialised response to the model."""
+        try:
+            op = self.operation
+            # build dict of raw encodable types U
+            raw_args = self.parameter_model.model_validate_json(json_str)
```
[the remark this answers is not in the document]
eb8680: Why don't we need an inverse of that? Is what comes back from `completion` already deserialized, as opposed to being an `OpenAIMessage`?

effectful/handlers/llm/encoding.py
kiranandcode: I haven't been able to successfully get any provider to return images at all. `openai` supports it, but only through the responses API. We don't have any usecases where we ask the llm to generate images, so I didn't spend too much time trying different models to work out if any supports image generation through what w ehave.

The main usecase is if we're supplying images as input to a template, or we have tools that return images. If these arguments are dataclasses which have images as fields, then this current functionality would make them work transparently, but given the increased maintenance burden/complexity, I can remove this.

effectful/handlers/llm/providers.py
kiranandcode: removed the comment, keeping the discussion on github.

effectful/handlers/llm/encoding.py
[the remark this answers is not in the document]
kiranandcode: done

effectful/handlers/llm/providers.py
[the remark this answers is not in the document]
kiranandcode: it's possible that serialize shouldn't be type directed. All of its behaviour is dynamic right now. We might want it to be extensible via defop, but after thinking about it more I think it should be a function that handles ImageUrl like dicts specially.

kiranandcode:
Test fails, as expected:
```
I retrieved a set of 5 images. Each image appears to be a simple placeholder graphic encoded in Base64. Here's a general description:

1. **Image 1 to 5:** All images are identical, consisting of a very basic visual placeholder. They are tiny (8x8 pixels) representations that lack distinctive features or detailed content, likely used just as a stand-in graphic, possibly for testing purposes. They don't contain any meaningful scenes or identifiable objects.

These are minimalistic graphics and don't offer much to describe. Please provide more specific criteria if you need a different type of image or additional detail!
```

effectful/handlers/llm/encoding.py
```diff
@@ -0,0 +1,261 @@
...
+    # Check if it's a dataclass and redirect to dataclass handler
+    if dataclasses.is_dataclass(ty):
+        return _type_encodable_type_dataclass(ty)
+    return _type_encodable_type_base(ty)
+
+
+@type_to_encodable_type.register(int)
+@type_to_encodable_type.register(float)
+@type_to_encodable_type.register(bool)
+@type_to_encodable_type.register(str)
```
datvo06: Is it possible, and should we make the default behavior identity for all classes, or for only a specific set of classes, we would define Encodable?

effectful/handlers/llm/providers.py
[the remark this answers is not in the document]
kiranandcode: fixed by making `serialize` a classmethod of encoded_ty

effectful/handlers/llm/providers.py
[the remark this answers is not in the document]
kiranandcode: handled in a more general and uniform way by moving `serialize` to be a class method of encodable.

PR #444, opened 2025-12-16
Class and Type Synthesis

### LINKED ISSUE #410: Support interface/class based program synthesis for the LLM decoding
Motivation from the Robotl repository. Beyond `Callable`'s, application domains often need to synthesize classes/objects with specific properties and constructors, such as:

```python
@multimodal_template(context=[Image.open("assets/specs/hammer1.png")])
class Hammer(Component):
    """Make a hammer that looks like the image"""  # The text-based specification goes here.

    height: int  # <--- spec attributes come from here, can be multiple attributes

    def __init__(
        self,
        width=5,
        height=10,
        handle_radius=1,
        handle_height=10,
        head_radius=1,
        head_height=10,
        **kwargs,
    ):
        pass
```

A clean interface to this would be allowing return types to be interfaces `ABC`s and then having the synthesis module return a class satisfying that type.

datvo06:
Closes #410, depends on #422

### DIFF
--- a/docs/source/agent.py
+++ b/docs/source/agent.py
@@ -1,34 +1,23 @@
 import functools
-from typing import Optional
 
 from effectful.handlers.llm import Template
-from effectful.handlers.llm.providers import compute_response, format_model_input
+from effectful.handlers.llm.providers import (
+    LiteLLMProvider,
+    compute_response,
+    format_model_input,
+)
 from effectful.ops.semantics import fwd, handler
 from effectful.ops.syntax import defop
+from effectful.ops.types import NotHandled
 
 
 class Agent:
-    '''When inheriting from Agent, Template-valued methods will have the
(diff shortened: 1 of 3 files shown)


effectful/handlers/llm/type_synthesis.py
```diff
@@ -0,0 +1,327 @@
...
+
+        return SynthesizedType(
+            type_name=type_name, module_code=textwrap.dedent(source).strip()
+        )
+
+    # Counter for unique filenames
+    _decode_counter: typing.ClassVar[int] = 0
+
+    @classmethod
+    def decode(cls, vl: SynthesizedType, context: LexicalContext | None = None) -> type:
```
eb8680: There's some really scary stuff happening in here. Is the sole aim to support calling `inspect.getsource` on the result of `decode`?

PR #456, opened 2025-12-22
Creating Encodable For Callable Type

datvo06:
Summarize changes:
1. Adding Optional `Template` parameter to decode
2. Use the `Template` to gather lexical context for decoding synthesized function
3. Simplify ProgramSynthesisHandler.


eb8680:
I think we can get rid of the handler/`get_synthesis_context` indirection once #498 and #499 are resolved.

PR #463, opened 2025-12-26
Fix method call templates

kiranandcode:
This PR fixes method call templates, closes #462 

A couple of changes:
```python
try:
  hints = get_type_hints(obj)
except Exception:
    ...
```
was always raising exceptions for Templates with the exception `inspect not found`. I could make it work by specifying `globalns=globals()` but then the returned hints dictionary was not the signature, but the signature of the template itself?

There is no reliable way to detect method templates, in this case, if the first parameter is named self, we drop it, thereby restoring ability for method calls to work.

PR #464, opened 2025-12-26
Remove `agent.py` from handler and add docs on how to implement Agents

### LINKED ISSUE #461: Remove `Agent` from `handlers.llm`
We're not sure that this is desirable enough to include in the library, but we should retain this code as an example.

kiranandcode:
This PR removes `agent.py` and updates `llm.ipynb` to describe how to implement Agents. Method templates were broken so it relies on #463  and should be merged after that lands. Closes #461

The agent class in `agent.py` also needed some modification to make it work with the latest functionality. It was actually somewhat broken.


kiranandcode:
will need to be cherry-picked on top of #424, changing to draft until that lands.

jfeser (changes requested):
That notebook is getting unwieldy, so let's add this to the documentation as a code example like the beam search.

jfeser (changes requested):

kiranandcode:
@eb8680 ah, good point, removed!

eb8680 (approved):

PR #467, opened 2025-12-26
Implemented record and replay fixtures for LLM calls

### LINKED ISSUE #458: LLM integration tests should be made more reliable
For example, `test_image_input` passes locally but is failing on #424. As far as I can tell, this doesn't indicate a bug.

A few options:
- We could run the tests repeatedly and pass if a sufficient fraction pass.
- We could trace a live LLM when writing the test and replay the trace when running it.

kiranandcode:
This PR modifies `test_handlers_llm.py` to implement record/replay fixtures. By default, when running locally `completion` calls are dumped to the folder `fixtures` with the test id that they are being run in.

When running in CI, the script will instead deserialize the fixtures instead of calling the LLM provider endpoint.

This should allow seamless testing of LLM behaviours locally, while making sure CI runs are deterministic.

Closes #458


tests/test_handlers_llm_provider.py
```diff
@@ -61,6 +69,25 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+class ReplayLiteLLMProvider(LiteLLMProvider):
+    @implements(completion)
+    def _completion(self, *args, **kwargs):
+        test_id = os.getenv("PYTEST_CURRENT_TEST", "unknown")
+        test_id = test_id.replace(" ", "_").replace("::", "__").replace("/", "_")
+        path = FIXTURE_DIR / f"{test_id}.json"
+        if IN_CI:
```
jfeser: I'd make the no-llm behavior the default both in and out of CI, and add a new makefile target to rebuild the snapshots.

jfeser (changes requested):

tests/test_handlers_llm_provider.py
kiranandcode: Fixed.

tests/test_handlers_llm_provider.py
```diff
@@ -61,6 +69,25 @@ def wrapper(*args, **kwargs):
     return decorator
 
 
+class ReplayLiteLLMProvider(LiteLLMProvider):
+    @implements(completion)
+    def _completion(self, *args, **kwargs):
+        test_id = os.getenv("PYTEST_CURRENT_TEST", "unknown")
```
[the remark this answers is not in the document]
kiranandcode: Done!

PR #474, opened 2025-12-29
Fail when encoding terms or operations

### LINKED ISSUE #471: `handlers.llm.encoding` should have failure cases for `Term` and `Operation`
We shouldn't allow encoding and decoding of the base types `effectful.ops.types.Term` and `effectful.ops.types.Operation`. These cases constitute undefined behavior and should fail by default; to support use cases like #430 that operate on subtypes of `Operation` we should have to add more specific overloads of `type_to_encodable_type`.

jfeser:
Closes #471

### DIFF
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -76,12 +76,12 @@ def _type_encodable_type_base[T](ty: type[T]) -> Encodable[T]:
 
 @type_to_encodable_type.register(Term)
 def _type_encodable_type_term[T: Term](ty: type[T]) -> Encodable[T]:
-    raise ValueError("Terms cannot be encoded or decoded in general.")
+    raise TypeError("Terms cannot be encoded or decoded in general.")
 
 
 @type_to_encodable_type.register(Operation)
 def _type_encodable_type_operation[T: Operation](ty: type[T]) -> Encodable[T]:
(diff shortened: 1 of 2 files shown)


effectful/handlers/llm/encoding.py
```diff
@@ -73,6 +74,16 @@ def decode(cls, vl: T) -> T:
...
 
 
+@type_to_encodable_type.register(Term)
+def _type_encodable_type_term[T: Term](ty: type[T]) -> Encodable[T]:
+    raise ValueError("Terms cannot be encoded or decoded in general.")
+
+
+@type_to_encodable_type.register(Operation)
+def _type_encodable_type_operation[T: Operation](ty: type[T]) -> Encodable[T]:
+    raise ValueError("Operations cannot be encoded or decoded in general.")
```
eb8680: nit: `ValueError` -> `TypeError`

tests/test_handlers_llm_encoding.py
```diff
@@ -6,6 +6,17 @@
 from PIL import Image
 
 from effectful.handlers.llm.encoding import type_to_encodable_type
+from effectful.ops.types import Operation, Term
+
+
+def test_type_to_encodable_type_term():
+    with pytest.raises(ValueError):
```
eb8680: nit: `ValueError` -> `TypeError`

eb8680 (approved):

PR #475, opened 2025-12-29
Remove program synthesis code

### LINKED ISSUE #470: `handlers.llm.synthesis` should be replaced with a stub in initial release
The synthesis machinery in `handlers.llm` is currently out of date and feature-incomplete following #424 and #434 and its design and implementation are going to evolve significantly over the next few weeks. Before we merge #363 into `master` we should replace it with a stub and mark all relevant tests as xfailing, and relocate development (including a number of current open PRs) to a new dedicated staging branch targeting `master`.

jfeser:
Closes #470 

To be restored once it's stabilized.

### DIFF
--- a/tests/test_handlers_llm_provider.py
+++ b/tests/test_handlers_llm_provider.py
@@ -239,6 +239,7 @@ class TestLLMLoggingHandler:
         )
 
 
+@pytest.mark.xfail(reason="Program synthesis not implemented")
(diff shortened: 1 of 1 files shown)


eb8680 (approved):
~~This seems fine but I don't think the test failure is random, can you take a look?~~

Edit: I was wrong, merging

PR #483, opened 2026-01-06
Fix RetryHandler not to use Template.replace()

### LINKED ISSUE #482: Template.replace()'s results can calls its former self.
When `Template.replace()` is called, it create a new Template and set the `__context__` to the original Template's context.
When `tools()` are called, `obj is self` is checked, since the new Template is not the original Template, it registers the old one.
Reproduction:
```python
"""Template defined in its own module."""

from pydantic import BaseModel
from effectful.handlers.llm import Template


class Output(BaseModel):
    value: str


@Template.define
def my_template(input: str) -> Output:
    """Process: {input}"""
    raise NotHandled
```
In another file
```python
from effectful.handlers.llm import Template
from my_template import my_template


if __name__ == "__main__":
    original = my_template
    
    print(f"Before replace:")
    print(f"  original.tools: {list(original.tools.keys())}")
    
    # Call Template.replace from THIS file (different from where my_template is defined)
    replaced = Template.replace(original, prompt_template="Modified: {input}")
    
    print(f"\nAfter replace:")
    print(f"  original.tools: {list(original.tools.keys())}")
    print(f"  replaced.tools: {list(replaced.tools.keys())}")
    
    # Assertions
    assert original.tools == {}, f"Original should have empty tools, got: {list(original.tools.keys())}"
    assert replaced.tools == {}, f"Replaced should have empty tools, got: {list(replaced.tools.keys())}"
    
    print("\n Both templates have empty tools")
```
Result:
```
Before replace:
  original.tools: []

After replace:
  original.tools: []
  replaced.tools: ['my_template']
Traceback (most recent call last):
  File "/Users/datnguyenthanh/Marc/OrchestSynth/scripts/issue_demo/test_replace_from_different_file.py", line 35, in <module>
    assert replaced.tools == {}, f"Replaced should have empty tools, got: {list(replaced.tools.keys())}"
           ^^^^^^^^^^^^^^^^^^^^
AssertionError: Replaced should have empty tools, got: ['my_template']
```

datvo06:
Partially closes #482


effectful/handlers/llm/providers.py
```diff
@@ -349,6 +311,61 @@ def format_model_input[**P, T](
...
+        if self._error_feedback:
+            messages = messages + [
+                {"type": "message", "content": self._error_feedback, "role": "user"}
+            ]
+        return messages
+
+    @implements(Template.__apply__)
+    def _retry_completion(self, template: Template, *args, **kwargs) -> Any:
+        """Retry template execution with error feedback injection."""
+        self._error_feedback = ""
```
eb8680: It's not correct to have this state live on `RetryLLMHandler` - nested `Template` calls with failures will overwrite each other. It should be in a separate handler like `InstructionsHandler` in #484.

datvo06: @eb8680 Thank you,
I see now. It might be the case that the error happens inside a nested template, so only the nested one would be corrected, but not the outer one get corrected. 
In that case, we might want to install an error feedback handler (`InstructionHandler`) specifically designed for the current scope. Because these handlers are installed according to the calling, they'd handle the right failure for the right template.
I'll implement this now.

PR #500, opened 2026-01-21
Adds context parameter to `encode` and `decode` calls (closes  #499)

### LINKED ISSUE #499: `Encodable` should take optional context arguments in `encode` and `decode`
To support advanced use cases like #497 we'll need the `Encodable` interface to be context-sensitive:
```py
class Encodable[T, U]:
  ...
  @classmethod
  def encode(cls, vl: T, env: Mapping[str, Any]) -> U: ...

  @classmethod
  def decode(cls, vl: U, env: Mapping[str, Any]) -> T: ...
```

kiranandcode:
This PR updates the Encodable class

```python
class Encodable[T, U]:
  ...
  @classmethod
  def encode(cls, vl: T, env: Mapping[str, Any]) -> U: ...

  @classmethod
  def decode(cls, vl: U, env: Mapping[str, Any]) -> T: ...
```

### DIFF
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -34,12 +34,12 @@ class EncodableAs[T, U](ABC):
 
     @classmethod
     @abstractmethod
-    def encode(cls, vl: T, env: Mapping[str, Any]) -> U:
+    def encode(cls, vl: T, env: Mapping[str, Any] | None = None) -> U:
         pass
 
     @classmethod
     @abstractmethod
-    def decode(cls, vl: U, env: Mapping[str, Any]) -> T:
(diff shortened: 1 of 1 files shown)


datvo06:
Based on our discussion yesterday, would it make sense to have another operation to retrieve context instead? Like in #456 ?

eb8680:
> would it make sense to have another operation to retrieve context instead?

I'm not sure I see a use case for this. What are the circumstances where would we need the context but not have access to it as a lexical variable?

eb8680 (commented):
Let's make the argument optional (`env: Mapping[str, Any] | None = None`), since a lot of the time it's not necessary.

Issue #490, opened 2026-01-19
Marginally less unsafe evaluation of LLM-generated code


eb8680:
When a `Template` generates higher-order objects like `Callable`s or `type`s that require some amount of Python evaluation to decode, we should do something a little less dangerous and error-prone than naively invoking Python's `builtins.eval`/`exec` on an arbitrary code string in the original application context.

There's no way to guarantee that any nontrivially expressive form of LLM-generated Python code execution is safe, but there are a few things we could look into that might help mitigate the worst failure modes mostly automatically:
- We temporarily removed the synthesis functionality in `handlers.llm` in part for this reason (#470). When it's added back evaluation should still only happen when a specific handler (or at least environment variable) is active, not by default.
- [`RestrictedPython`](https://github.com/zopefoundation/RestrictedPython) implements a more comprehensive and configurable suite of analyses and transformations at the level of `ast.AST` objects as well as drop-in replacements for `builtins.compile`/`exec`/`eval` aimed at preventing code from accessing the surrounding Python environment. We could potentially use one or both parts of this library directly.
- [`smolagents`](https://huggingface.co/docs/smolagents/index) implements a [meta-circular evaluator](https://github.com/huggingface/smolagents/blob/main/src/smolagents/local_python_executor.py#L1411) of Python `ast.AST`s for a subset of valid Python code. Using or adapting that code would be straightforward and it could have applications beyond the LLM module. A related idea would be to implement a custom interpreter at the level of CPython VM bytecode rather than `ast.AST`, as in #288 
- Type annotations (#451), doctests (#433) and string formatting may sneakily trigger calls to `builtins.eval` even inside a meta-circular evaluator. We might want to do something about these edge cases specifically, e.g. scrubbing annotations from generated code after typechecking.
- [`progent`](https://arxiv.org/abs/2504.11703) implements a tiny DSL for tool access policies that is not unrelated to our discussion of effect types in #448 
- All of the above are distinct from and complementary to sandboxing, where we run a Python process or subprocess in a fresh, isolated OS-level environment (e.g. a Docker container) or even a remote machine (e.g. AWS Lambda). Probably any deployed LLM application that is both connected to the outside world and executing generated code should be running in a container, something we could attempt to heuristically detect and warn a user about. However, deeply nesting sandboxes at the level of individual `Template` calls seems clunky and challenging to debug without much additional security upside.

[...]

PR #503, opened 2026-01-22
Add small restricted meta circular interpreter

kiranandcode:
As part of work towards #490, this (draft) PR adds a small meta-circular interpreter with tests to the effectful library. Leaving as a draft as we may want to discuss the interface a bit and figure out how to integrate it with the LLM code generation.

The tests are pretty exhaustive, and include an example of using the meta-circular interpreter to evaluate itself and then running tests, which seems to make it very likely that the code is mostly correct.

### DIFF
--- a/effectful/internals/meta_eval.py
+++ b/effectful/internals/meta_eval.py
@@ -1,0 +1,2359 @@
+import ast
+import builtins
+import inspect
+import linecache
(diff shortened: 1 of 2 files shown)


eb8680:
Closing as not planned for now, although we may end up revisiting this later

PR #518, opened 2026-01-28
update `encoding_instructions` to take in lexical context as a parameter

### LINKED ISSUE #516: `encoding_instructions` should take lexical context
Generalisation of the pydantic json schema that is sent to the llm
```python
class Encodable[T](EncodableAs[T, type]):
  ...
  @Operation.define  # or only define an `Operation` for specific `Encodable` subtypes?
  @classmethod
  def encoding_instructions(cls, env: Mapping[str, Any]) -> Sequence[str]: ...
```

kiranandcode:
Leaving making it into an operation into another PR. Closes #516

### DIFF
--- a/tests/test_handlers_llm_encoding.py
+++ b/tests/test_handlers_llm_encoding.py
@@ -718,3 +718,29 @@ def test_type_to_encodable_type_nested_pydantic_model():
     assert decoded_from_model == person
     assert isinstance(decoded_from_model, Person)
     assert isinstance(decoded_from_model.address, Address)
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


kiranandcode:
> After reflecting on https://github.com/BasisResearch/effectful/pull/519 I wonder if @jfeser's other idea to fold this information into Encodable.t rather than making it a separate piece of Encodable is the more sensible approach.

could you give a code snippet to clarify? I'm still not entirely sure I have a good idea of what this idea is. As in t should be a type that supports .encoding_instructions?

kiranandcode:
Subsumed by #521

Issue #515, opened 2026-01-27
RestrictedPythonEval handler for program synthesis


kiranandcode:
Provide a handler for compile and exec that uses Restricted Python: https://restrictedpython.readthedocs.io/en/latest/usage/basic_usage.html.

PR #520, opened 2026-01-28
EvalProvider using `RestrictedPython` [Depends on #519]

kiranandcode:
Building on top of #519, this provides another handler for `parse`, `eval`, `compile` that calls out to `RestrictedPython` instead to provide some safety. Closing #515,  The core of this implementation is:

```python
class RestrictedEvalProvider(ObjectInterpretation):
     ...
    @implements(compile)
    def compile(self, module: ast.Module, filename: str) -> CodeType:
        return compile_restricted(module, filename=filename, mode="exec")

    @implements(exec)
    def exec(self, bytecode: CodeType, env: dict[str, Any]) -> None:
        rglobals = safe_globals.copy()
        for k, v in env.items():
            if k != "__builtins__":
                rglobals[k] = v
        builtins.exec(bytecode, rglobals, env)
```

`parse` is the same as `UnsafeEval`. The tests have been updated to be parametric in the eval provider and test RestrictedEval as well.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,9 +28,7 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
-dependencies = [
-    "restrictedpython>=8.1",
-]
+dependencies = []
 
 [project.urls]
 Homepage = "https://www.basis.ai/"
@@ -46,6 +44,7 @@ llm = [
   "litellm",
   "pillow",
   "pydantic",
+  "restrictedpython>=8.1"
 ]
 prettyprinter = ["prettyprinter"]
 docs = [


effectful/handlers/llm/evaluation.py
```diff
@@ -0,0 +1,153 @@
...
+        rglobals.update({k: v for k, v in env.items() if k != "__builtins__"})
+
+        # Enable for loops and comprehensions
+        rglobals["_getiter_"] = Eval.default_guarded_getiter
+
+        # Enable sequence unpacking in comprehensions and for loops
+        rglobals["_iter_unpack_sequence_"] = Guards.guarded_iter_unpack_sequence
+        rglobals["getattr"] = Guards.safer_getattr
+        rglobals["setattr"] = Guards.guarded_setattr
+        rglobals["_write_"] = lambda x: x
```
eb8680: A lot of this seems like stuff that should be determined from arguments passed to the `RestrictedEvalProvider` constructor, rather than hardcoded in `exec`.

effectful/handlers/llm/evaluation.py
```diff
@@ -0,0 +1,153 @@
...
+            None,
+            source.splitlines(True),
+            filename,
+        )
+        return ast.parse(source, filename=filename, mode="exec")
+
+    @implements(compile)
+    def compile(self, module: ast.Module, filename: str) -> CodeType:
+        # RestrictedPython can compile from an AST directly.
+        return compile_restricted(module, filename=filename, mode="exec", **self.config)
```
eb8680: Are there not limits on what `self.config` can contain imposed by `compile_restricted`? What keyword arguments does `compile_restricted` actually take?

effectful/handlers/llm/evaluation.py
kiranandcode: or do you mean we should take in additional flags to indicate which features to support?

kiranandcode: Updated.

effectful/handlers/llm/evaluation.py
kiranandcode: Updated. The main parameter compile_restricted takes is a policy.

effectful/handlers/llm/evaluation.py
```diff
@@ -0,0 +1,210 @@
...
+        # Build restricted globals from RestrictedPython's defaults
+        rglobals: dict[str, Any] = safe_globals.copy()
+
+        if self.allow_class_definitions:
+            # Enable class definitions (required for Python 3)
+            rglobals["__metaclass__"] = type
+            rglobals["__name__"] = self.module_name
+
+        # Layer `env` on top (without letting callers replace the restricted builtins).
+        rglobals.update({k: v for k, v in env.items() if k != "__builtins__"})
```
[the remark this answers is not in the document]
kiranandcode: passing different objects for locals and globals causes exec to behave differently - it treats it as within a class definition which isn't the semantics we want (we want a toplevel module).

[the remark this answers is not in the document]
kiranandcode: ah, good point, will add a test.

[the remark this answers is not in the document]
kiranandcode: Added

pyproject.toml
```diff
@@ -28,7 +28,9 @@ classifiers = [
     "Programming Language :: Python :: 3.12",
     "Programming Language :: Python :: 3.13",
 ]
-dependencies = []
+dependencies = [
+    "restrictedpython>=8.1",
```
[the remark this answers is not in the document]
kiranandcode: ah, yep will update

PR #527, opened 2026-02-02
Update internal interface to enforce message ids

kiranandcode:
This PR updates the internals of `completions.py` such that messages can be assumed to have `id`s, which will be useful for caching.


eb8680 (commented):
Do we need to strip the `id` field from the `Message` before passing it to `completion`, or do the underlying LLM APIs not care?

eb8680 (approved):

PR #528, opened 2026-02-02
Abstract message sequence to handlers that can be overriden.

kiranandcode:
This PR updates LiteLLMProvider to handle message sequences through a handler. The `agent.py` has been updated to reuse this mechanism. The tests have been updated to enforce the respective invariants for `call_tool` and `call_assistant`, that is:

call_tool: fresh message sequence
call_assistant: does not duplicate messages, only saves messages on successful execution of fwd

### DIFF
--- a/docs/source/agent.py
+++ b/docs/source/agent.py
@@ -5,7 +5,7 @@ from effectful.handlers.llm import Template
 from effectful.handlers.llm.completions import (
     LiteLLMProvider,
     Message,
-    MessageSequence,
+    get_message_sequence,
 )
 from effectful.ops.semantics import handler
 from effectful.ops.types import NotHandled
@@ -25,7 +25,7 @@ class Agent:
 
             @functools.wraps(template)
             def wrapper(self, *args, **kwargs):
-                with handler(MessageSequence(self.state)):
+                with handler({get_message_sequence: lambda: self.__history__}):
                     return template(self, *args, **kwargs)
 
             setattr(cls, method_name, wrapper)
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -26,40 +26,46 @@ from litellm import (
 from effectful.handlers.llm.encoding import Encodable
 from effectful.handlers.llm.template import Template, Tool
 from effectful.ops.semantics import fwd, handler
-from effectful.ops.syntax import ObjectInterpretation, implements
+from effectful.ops.syntax import ObjectInterpretation, defop, implements
 from effectful.ops.types import Operation
 
 
-class MessageID(typing.TypedDict, total=False):
+class AssistantMessage(OpenAIChatCompletionAssistantMessage):
     id: str
 
 
-class AssistantMessage(OpenAIChatCompletionAssistantMessage, MessageID):
-    pass
+class ToolMessage(ChatCompletionToolMessage):
+    id: str
 
 
-class ToolMessage(ChatCompletionToolMessage, MessageID):
-    pass
+class FunctionMessage(ChatCompletionFunctionMessage):
+    id: str
 
 
-class FunctionMessage(ChatCompletionFunctionMessage, MessageID):
-    pass
+class SystemMessage(OpenAIChatCompletionSystemMessage):
+    id: str
 
 
-class SystemMessage(OpenAIChatCompletionSystemMessage, MessageID):
-    pass
-
-
-class UserMessage(OpenAIChatCompletionUserMessage, MessageID):
-    pass
+class UserMessage(OpenAIChatCompletionUserMessage):
+    id: str
 
 
 Message = AssistantMessage | ToolMessage | FunctionMessage | SystemMessage | UserMessage
 
 
+@defop
+def get_message_sequence() -> collections.OrderedDict[str, Message]:
+    return collections.OrderedDict()
+
+
+def append_message(message: Message):
+    get_message_sequence()[message["id"]] = message
+
+
 def _make_message(content: dict) -> Message:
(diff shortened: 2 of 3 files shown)


kiranandcode:
Depends on #527

docs/source/agent.py
```diff
@@ -32,28 +34,26 @@ def wrapper(self, *args, **kwargs):
...
-            prompt = self.state
-        return prompt
-
-    def _compute_response(self, *args, **kwargs):
-        # save response into persisted state
-        response: Message = fwd()
-        if Agent.current_agent() is self:
-            self.state.append(response)
-        return response
+    def _call_assistant(self, messages: list[Message], *args, **kwargs):
```
eb8680: I'd expect this to go away entirely

docs/source/agent.py
```diff
@@ -32,28 +34,26 @@ def wrapper(self, *args, **kwargs):
                 with handler(
                     {
                         Agent.current_agent: lambda: self,
-                        call_user: self._format_model_input,
-                        call_assistant: self._compute_response,
+                        call_assistant: self._call_assistant,
```
eb8680: I'd expect this to simplify to `handler(MessageSequence(self.state))`, we should no longer need `Agent.current_agent` or `Agent._call_assistant` and `Agent` should reduce to a tiny syntactic wrapper that installs the new handler around `Template` methods.

effectful/handlers/llm/completions.py
```diff
@@ -474,7 +518,7 @@ def __init__(self, model="gpt-4o", **config):
     def _call[**P, T](
         self, template: Template[P, T], *args: P.args, **kwargs: P.kwargs
     ) -> T:
-        messages: list[Message] = [*call_system(template)]
+        self.message_sequence_stack.append(collections.OrderedDict())
```
eb8680: This feels like it should be a handler context:
```py
with handler(MessageSequence()):
  ...  # body of _call
```

eb8680 (commented):
Let's see if we can simplify this a bit

effectful/handlers/llm/completions.py
```diff
@@ -443,12 +459,56 @@ def _call_tool(self, tool_call: DecodedToolCall) -> Message:
...
+
+        frame = self.message_sequence_stack[-1]
+        prefix = [m for msg_id, m in frame.items() if msg_id not in seen_ids]
+
+        message, tool_calls, result = fwd(prefix + msg_list, *args, **kwargs)
+        frame[message["id"]] = message
+        return message, tool_calls, result
+
+
+class LiteLLMProvider(LLMMessageSequenceProvider):
```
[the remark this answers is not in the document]
kiranandcode: If we separate it out, does this mean we'd require clients to install both?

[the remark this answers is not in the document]
eb8680: I'd expect it to be primarily used internally - it would be installed within the body of `LiteLLMProvider`'s `Template.__apply__` handler, or through an interface like the `Agent` class in the example.

effectful/handlers/llm/completions.py
kiranandcode: Updated

effectful/handlers/llm/completions.py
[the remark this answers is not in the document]
kiranandcode: Updated

docs/source/agent.py
kiranandcode: Simplified!

effectful/handlers/llm/completions.py
```diff
@@ -459,12 +459,60 @@ def _call_tool(self, tool_call: DecodedToolCall) -> Message:
...
+        return message, tool_calls, result
+
+
 class LiteLLMProvider(ObjectInterpretation):
     """Implements templates using the LiteLLM API."""
 
     config: collections.abc.Mapping[str, typing.Any]
 
     def __init__(self, model="gpt-4o", **config):
+        super().__init__()
```
eb8680: Remove

docs/source/agent.py
eb8680: Now dead code, should be removed

effectful/handlers/llm/completions.py
```diff
@@ -474,35 +522,32 @@ def __init__(self, model="gpt-4o", **config):
...
+            env = template.__context__.new_child(bound_args.arguments)
+
+            # Create response_model with env so tools passed as arguments are available
+            response_model = Encodable.define(
+                template.__signature__.return_annotation, env
+            )
 
-        # Create response_model with env so tools passed as arguments are available
-        response_model = Encodable.define(template.__signature__.return_annotation, env)
+            user_messages: list[Message] = call_user(template.__prompt_template__, env)
```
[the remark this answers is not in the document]
kiranandcode: Updated!

kiranandcode:
> Another invariant we should be enforcing and testing is that Messages are always ordered correctly in the input to completion. I'm a little concerned that this will reorder histories by ordering Messages by handler context, rather than actual global time.

yes, that is a problem,

```python
        msg_list = list(messages)
        seen_ids = {m["id"] for m in msg_list}

        prefix = [
            m for msg_id, m in self.message_sequence.items() if msg_id not in seen_ids
        ]

        message, tool_calls, result = fwd(prefix + msg_list, *args, **kwargs)
        self.message_sequence[message["id"]] = message
        return message, tool_calls, result
```
This implementation of `MessageSequence.call_assistant` was specifically written because messages were not being inserted in the right order. Maybe messagesequence shouldn't be an ordered dict, but a dict, and message_sequence should order by id, or timestamp?

eb8680 (approved):

PR #532, opened 2026-02-03
Fix retry llm handler swallowing tool call in message

### LINKED ISSUE #531: `RetryLLMHandler` drops tool calls
Retry LLM Handler drops tool calls.

```python
class RetryLLMHandler(ObjectInterpretation):

    @implements(call_assistant)
    def _call_assistant[T, U](
        self,
        tools: collections.abc.Mapping[str, Tool],
        response_format: Encodable[T, U],
        model: str,
        **kwargs,
    ) -> MessageResult[T]:
        message_sequence = get_message_sequence().copy()
        last_attempt = self.num_retries

        for attempt in range(self.num_retries + 1):
            try:
                # call assistant, use saved message_sequence
                with handler({get_message_sequence: lambda: message_sequence}):
                    message, tool_calls, result = fwd(
                        tools, response_format, model, **kwargs
                    )

                # ... need to reinstall tool call messages and results here
                append_message(message)
                return (message, tool_calls, result)

            except (ToolCallDecodingError, ResultDecodingError) as e:
                # On last attempt, re-raise to preserve full traceback
                if attempt == last_attempt:
                    raise

    @implements(call_tool)
    def _call_tool(self, tool_call: DecodedToolCall) -> Message:
        """Handle tool execution with runtime error capture.

        Runtime errors from tool execution are captured and returned as
        error messages to the LLM. Only exceptions matching `catch_tool_errors`
        are caught; others propagate up.
        """
        try:
            message = fwd(tool_call)
            # store successful messages?
            return message
        except self.catch_tool_errors as e:
            error = ToolCallExecutionError(tool_call.tool.__name__, tool_call.id, e)
            return error.to_feedback_message(self.include_traceback)

### PR DESCRIPTION
**Pull request by datvo06**
closes #531

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -475,8 +475,7 @@ class RetryLLMHandler(ObjectInterpretation):
         are caught; others propagate up.
         """
         try:
-            message = fwd(tool_call)
-            return message
+            return fwd(tool_call)
         except self.catch_tool_errors as e:
             error = ToolCallExecutionError(tool_call.tool.__name__, tool_call.id, e)
             message = error.to_feedback_message(self.include_traceback)

### CODE REVIEW

File: effectful/handlers/llm/completions.py [hunk 1]
```diff
@@ -464,7 +470,7 @@ def _call_assistant[T, U](
     @implements(completion)
     def _completion(self, *args, **kwargs) -> typing.Any:
         """Inject num_retries for litellm's built-in network error handling."""
-        return fwd(*args, num_retries=self.num_retries, **kwargs)
+        return fwd(*args, **{**kwargs, "num_retries": self.num_retries})
```
kiranandcode: `**({"num_retries": self.num_retries} | kwargs)`?

PR #544, opened 2026-02-06
Integrating Doctest

### LINKED ISSUE #433: Support and exploit doctests in higher-order `Template` specifications
Python's [`doctest`](https://docs.python.org/3/library/doctest.html) library allows classes, functions or modules to include runnable test code in their docstrings that can be accessed and evaluated programmatically. We should allow synthesis agents to exploit that as an additional source of feedback - they should be able to write docstrings that include doctests for the code they generate, and to run the doctests to see if their code works in the same way they might run the typechecker (#361).

datvo06:
Attempting to close #433 . Current implementation captures the template docstring at `__apply__` time and add it to the synthesized function's docstring and run it.


eb8680 (commented):
This is neat, but I'm not sure it makes sense to use doctests in the way that's proposed in this PR (or even my original issue #433), because it departs from the semantics they have for ordinary Python functions.

Instead, there are two cases where it would be more natural to incorporate doctests as semantic constraints on `Template`s, which may only be useful if the doctests are stripped from the prompt during the constraint satisfaction step so that it cannot simply memorize them:
1. When implementing `Template`s using ordinary tool-calling, the doctests should induce a prefix to the conversation in which the LLM learns to satisfy the semantic constraints in the doctests. This only needs to happen once per `Template` definition, rather than once per call, and should be thought of as augmenting the system prompt associated with that `Template`.
2. When generating and executing code to implement a `Template` (as in #526), the generated code should be required to pass the doctests to be valid. This needs to happen once per code generation step, which could either happen once per `Template` definition or once per `Template` call.

effectful/handlers/llm/doctest.py
```diff
@@ -0,0 +1,314 @@
...
+        invoked with the example's arguments (prompt stripped of doctests).
+        All conversation turns — including any incorrect attempts — are
+        accumulated into a prefix that is cached for future calls, so the
+        LLM can learn from the full experience.
+        """
+        prefix_messages: list[Message] = []
+        self._calibrating.add(template)
+
+        try:
+            for example in examples:
```
datvo06: This is one place where I'm unhappy with the current implementation: We have doctest and we have retry handler, it should have been cleaner to collect the test results.

eb8680:
Closing in favor of #694

PR #547, opened 2026-02-08
Add simple LLM examples

eb8680:
Depends on #545 #546 

This exploratory PR updates and extends @kiranandcode's old [Pocketflow example translations](https://github.com/BasisResearch/effectful/blob/pocketflow-examples-llm/docs/source/pocketflow-examples.ipynb) to work with the updated LLM module, and adds more translated from Pocketflow and elsewhere, including two versions of #404 

The code was mostly written by Claude, but I've reviewed, tweaked and executed most of it and verified that most of the example scripts seems to behave correctly even with less capable local LLMs. I don't think these are necessarily the most compelling examples (e.g. there is a lot of duplication, and a lot that is subsumed by a generic tool-calling agent loop), and I'm not sure we want to land them on `master`, but it's an interesting set of design patterns to consider.


eb8680:
Closing in favor of #694

PR #548, opened 2026-02-09
Move tool/tool call encoding logic from completion to encoding

eb8680:
This refactoring PR cleans up some logic in `handlers.llm.completions` related to tool call encoding and decoding by coercing it into the existing `Encodable` API.

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -228,8 +228,9 @@ def call_assistant[T, U](
             "final response from the model should be a string"
         )
         try:
-            raw_result = response_format.deserialize(serialized_result)
-            result = response_format.decode(raw_result)
+            result = response_format.decode(
+                response_format.deserialize(serialized_result)
+            )
         except (pydantic.ValidationError, TypeError, ValueError, SyntaxError) as e:
             raise ResultDecodingError(e, raw_message=raw_message) from e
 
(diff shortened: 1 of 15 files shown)


effectful/handlers/llm/completions.py
```diff
@@ -264,16 +186,15 @@ def call_assistant[T, U](
...
+    tool_specs = {
+        k: Encodable.define(type(t), tools).encode(t)  # type: ignore
+        for k, t in tools.items()
+    }
     messages = list(_get_history().values())
     response: litellm.types.utils.ModelResponse = completion(
         model,
         messages=list(messages),
-        response_format=response_model,
+        response_format=None if response_format.enc is str else response_format.enc,
```
datvo06: `response_model`  is changed to `response_format.enc`.
`response_format.enc` is just a python type, not asserted to be a Pydantic model (created through `pydantic.create_model`).
I understand that we want to remove the extra `value` wrapping layer.
But `_encodable_object` is not handling dataclass: `enc = ty` sets the type to raw class.
I think I can add a 1-line fix inside there and see if it works.

datvo06: Also, because we are not creating pydantic model here anymore, and instead leave that to encodable, we would need to handle primitive types as well.

datvo06 (changes requested):
I have one concern with using `enc` directly. It's ok, but we have to make the encoding complete.

datvo06:
I accidentally pushed the fix. I reverted it and making a PR on top.

PR #551, opened 2026-02-10
Validating docstring and allow referring to symbols in lexical context in the docstring

### LINKED ISSUE #459: `Template` constructor should validate docstring formatting
When constructing `Template`s, we should check that the variables in the docstring match the `Template` signature. The string should be a valid [format string](https://docs.python.org/3/library/string.html#format-string-syntax) that refers to valid lexical variables in the `Template` body's lexical scope.

For example, the following should always be disallowed when `author` is undefined:
```py
@Template.define
def write_poem(topic: str) -> str:
    """Write a poem about {topic} by {author}"""
    raise NotHandled
```
When `author` is a lexical variable in an enclosing scope, we could allow it to be referenced, or we could be strict and require `Template`s to refer only to arguments:
```py
def poet(author: str) -> Template[[str], str]:
    @Template.define
    def write_poem(topic: str) -> str:
        """Write a poem about {topic} by {author}"""
        raise NotHandled
    return write_poem
```

These checks could be implemented by attempting to parse the template according to the [official format string mini-language specification](https://docs.python.org/3/library/string.html#format-specification-mini-language), and by attempting to substitute dummy values into variable names via `.format(**self.__signature__.bind(*dummy_args, **dummy_kwargs).arguments)`

Note: Python 3.14 adds a new module [`string.templatelib`](https://docs.python.org/3/library/string.templatelib.html#string.templatelib.Template) with a richer [backwards-incompatible template string semantics](https://peps.python.org/pep-0750/); if we choose to support that in our prompt templates, which is out of scope for this issue, we should do similar validation.

datvo06:
Attempting to close #459. 
```python
@Template.define
def write_poem(topic: str) -> str:
    """Write a poem about {topic}"""
    raise NotHandled

def poet(author: str) -> Template[[str], str]:
    @Template.define
    def write_poem(topic: str) -> str:
        """Write a poem about {topic} by {author}"""
        raise NotHandled
    return write_poem
```

### DIFF
--- a/effectful/handlers/llm/template.py
+++ b/effectful/handlers/llm/template.py
@@ -327,10 +327,8 @@ class Template[**P, T](Tool[P, T]):
         op = super().define(default, *args, **kwargs)
         op.__context__ = context  # type: ignore[attr-defined]
 
-        # Operation.__get__ re-wraps templates as bound MethodType callables; validating
-        # those bound wrappers drops `self` from the signature and can falsely reject
-        # valid prompt fields like `{self.name}`. Keep validation on original define-time
-        # callables, but skip the bound wrapper path.
+        # Keep validation on original define-time callables, but skip the bound wrapper path.
+        # to avoid dropping `self` from the signature and falsely rejecting valid prompt fields like `{self.name}`.
         is_bound_wrapper = (
             isinstance(default, types.MethodType) and default.__self__ is not None
         )


eb8680 (dismissed):

effectful/handlers/llm/template.py
```diff
@@ -52,6 +54,52 @@ def infer_annotations(cls, sig: inspect.Signature) -> inspect.Signature:
...
+    """Validate that all format string variables in the template docstring
+    refer to names that can be resolved at call time.
+
+    Each variable must be either a parameter in the template's signature
+    or a name captured in the template's lexical context.
+
+    :raises ValueError: If any format string variable cannot be resolved.
+    """
+    doc = op.__default__.__doc__
+    if not doc:
```
eb8680: `Template` should always have a docstring

datvo06:
I updated the change.

Question - Should we support staticmethod validation?
Two new tests are currently be marked as skipping:
```python
@pytest.mark.xfail(
    reason="staticmethod + lexical scope: _define_staticmethod re-enters "
    "Template.define, causing frame walking to capture wrong context"
)
def test_validate_staticmethod_lexical_scope():
    """Staticmethod templates should capture lexical scope variables."""
    feet_per_mile = 5280  # noqa: F841

    @Template.define
    @staticmethod
    def convert(feet: int) -> str:
        """How many miles is {feet} feet? There are {feet_per_mile} feet per mile."""
        raise NotHandled

    # The inner template should have the correct lexical context
    inner = convert.__func__
    assert "feet_per_mile" in inner.__context__


@pytest.mark.xfail(
    reason="staticmethod + lexical scope: _define_staticmethod re-enters "
    "Template.define, causing frame walking to capture wrong context"
)
def test_staticmethod_lexical_scope_formatting():
    """Staticmethod templates should format lexical scope variables at runtime."""
    feet_per_mile = 5280  # noqa: F841

    @Template.define
    @staticmethod
    def convert(feet: int) -> str:
        """How many miles is {feet} feet? There are {feet_per_mile} feet per mile."""
        raise NotHandled

    with handler(TemplateStringIntp()):
        assert (
            convert(7920)
            == "How many miles is 7920 feet? There are 5280 feet per mile."
        )
```
A patch that would pass this is:
```python
        if isinstance(default, staticmethod):
            # Note: Handle staticmethod explicitly to avoid re-entering Template.define and its frame walking.
            # See: test_validate_staticmethod_lexical_scope
            inner = typing.cast(
                Template[Q, V],
                super().define(default.__func__, *args, **kwargs),
            )
            inner.__context__ = context
            cls._validate_prompt(inner, context)
            return typing.cast(Template[Q, V], staticmethod(inner))

        op = typing.cast(
            Template[Q, V], super().define(default, *args, **kwargs)
        )
        op.__context__ = context
        cls._validate_prompt(op, context)
        return op
```

PR #563, opened 2026-02-13
Use `tenacity` to implement more general `RetryLLMHandler`

eb8680:
Depends on #562 

Motivated by the need to add `tenacity` as a dependency in https://github.com/BasisResearch/RoboTL/pull/372, this PR adds `tenacity` as a dependency to `effectful[llm]` and reimplements `RetryLLMHandler` using the more general functionality available there.

### DIFF
--- a/docs/source/llm.ipynb
+++ b/docs/source/llm.ipynb
@@ -16,7 +16,7 @@
   },
   {
    "cell_type": "code",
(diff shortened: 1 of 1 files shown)


datvo06:
It works great! I'm re-running the notebook to get the full output.

Issue #489, opened 2026-01-19
`Tool` and `Template` signatures should support polymorphic types


eb8680:
We should take advantage of the type inference engine in `internals/unification.py` to allow `Template`s and `Tool`s to be polymorphic:
```py
@Template.define
def generate_similar[T](examples: Sequence[T]) -> T:
  """Generate another instance of the items in {examples}"""

...

@Tool.define
def extend_sequence[T](examples: Sequence[T], new_example: T) -> Sequence[T]:
  """Extends the input sequence with a new example"""
  ...
```
Here `generate_similar` can infer its type variable and enforce the resulting concrete type constraint, and `extend_sequence` can similarly have the type constraint on its arguments enforced immediately when a tool call to `extend_sequence` is generated.

datvo06:
Happy to take this as I'm working on RoboTL#444 and RoboTL#445

eb8680:
@datvo06 this may require some discussion, see @kiranandcode's comment above about the ambiguity that has to be resolved by an LLM calling a polymorphic tool.

datvo06:
Right. I was drafting a plan, and that was one headache. I was thinking that it should be possible to remove the ambiguity by installing a context manager that keeps track of the call stack and the type.
Concretely, we add this ops:
```python
@defop
def _get_active_type_subs() -> "Substitutions":
    """Effect operation: returns the active TypeVar substitution for the
    enclosing Template call, or {} if no Template call is in flight.
    Installed by LiteLLMProvider._call for the duration of each call.
    """
    return {}
```

If we install the handler at `_call_assistant` time, then we would have enough context to decode it. The Tool spec sent to OpenAI is Example[int] (or whatever), and the LLM is constrained to produce the right shape.
We can also use this for decoding.

eb8680:
I implemented code-mode tool calling in #694

eb8680:
Resolved in #694 via typechecked code-mode tool calling

PR #570, opened 2026-02-16
Add support for final-answer tool calls

### LINKED ISSUE #549: `Template`s should be able to return `Tool` call results as final answers
Sometimes we may want `Template`s to return the result of a `Tool` call directly as a final answer, rather than always passing through a final LLM API call. This includes #526 as a special case, as well as settings where we want `Template`s to call out to other generative models (e.g. image or video generation) whose outputs may not be compatible with a provider.

eb8680:
Addresses #549 

This PR adds a new `effectful.ops.types.Annotation` for `Tool` return types, `effectful.handlers.llm.template.IsFinal`, such that when a `Template` generates a call to an `IsFinal`-annotated `Tool` whose return type is compatible with that of the `Template`, the result of the tool call will be returned directly from `call_assistant` and `Template.__apply__` as the final result of the `Template` call with no further `call_assistant` turns or serialization/postprocessing.

This PR corresponds to one fairly conservative corner of the design space sketched in #549:
- Only well-formed and successful tool calls can be used as final answers - if an LLM-generated `IsFinal` tool call cannot be decoded with valid arguments and a static output type that matches the original `Template`'s, or if executing the tool call triggers an error at runtime that is captured by `RetryLLMHandler`, the `IsFinal` annotation will be disregarded and the LLM will proceed to another `call_assistant` step as usual.
- it does not allow a `Template` to use arbitrary type-compatible but un-annotated `Tool`s to get a final answer - if a `Tool` is not an explicitly `IsFinal`-annotated final-answer tool, its output will always be sent through at least one more `call_assistant` step.
- It does not allow a `Template` to use an `IsFinal`-annotated `Tool` with incompatible return type as a non-final tool; attempting to call such a tool will trigger an error.
- It does not *require* the LLM to always generate an `IsFinal` tool call to compute a final answer when at least one type-compatible `IsFinal`-annotated `Tool` is available - the LLM can use different tools or even none at all to get a result.

Any or all of these points might be things to consider changing prior to landing this PR - I would expect the functionality in #549 and especially this instantiation to be useful mostly in cases like #526 where we have something very general like a code-generation or text-to-image-generation tool which can be used with many `Template`s and which we always want to use whenever it is type-compatible.

This would probably be much easier to use in conjunction with polymorphism #489

### DIFF
--- a/effectful/handlers/llm/__init__.py
+++ b/effectful/handlers/llm/__init__.py
@@ -1,3 +1,3 @@
-from .template import Agent, IsFinalAnswer, Template, Tool
+from .template import Agent, Template, Tool
 
-__all__ = ["Agent", "IsFinalAnswer", "Template", "Tool"]
+__all__ = ["Agent", "Template", "Tool"]
--- a/effectful/handlers/llm/__init__.py
+++ b/effectful/handlers/llm/__init__.py
@@ -1,3 +1,3 @@
-from .template import Agent, Template, Tool
+from .template import Agent, IsFinalAnswer, Template, Tool
 
-__all__ = ["Agent", "Template", "Tool"]
+__all__ = ["Agent", "IsFinalAnswer", "Template", "Tool"]
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -319,7 +319,7 @@ def call_assistant[T, U](
         tool_ret = inspect.signature(final_tc.tool).return_annotation
         if typing.get_origin(tool_ret) is typing.Annotated:
             tool_ret = typing.get_args(tool_ret)[0]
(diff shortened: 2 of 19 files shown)


datvo06 (changes requested):
At the first glance, I think the check in `call_assistant` can cause some troubles (below). I don't have a clear answer on how to address them yet, but we can add tests, mark xfail and create issues.

eb8680:
> I'm skeptical that tool calling training will allow LLMs to effectively use tools that have the side effect of ending the interaction.

Isn't this just a funny sort of structured output, followed by extra information from the tool call in the next user message? It's possible that it doesn't work that well in practice but I know it's a pattern that's used quite a bit in `smolagents`, among other libraries.

Issue #553, opened 2026-02-11
Encodable implementations should be tested more thoroughly


eb8680:
It shouldn't be possible for bugs like #552 to slip past unit tests and land on `master`. We should have much more extensive unit tests for our `Encodable` implementations.

PR #574, opened 2026-02-17
Refactor encodable tests to be more comphrehensive

kiranandcode:
This PR updates `test_handlers_llm_encoding.py` to be more clean and comprehensive.

It first replaces the bespoke `test_...` functions with a single parameterised test function that runs encode and decode.

It also adds a randomised test suite of randomly generated structured types (tuples, lists, named tuples, pydantic models) and callables over those types and checks that the encode,decode pipeline works with those.

For the cases that `Encodable` does not yet support and we have pending PRs, it includes tests for them them but marks them as `skipif`:

- variable length tuples

```python
    pytest.param(
        tuple[int, ...],
        (1, 2, 3),
        id="tuple[int,...]",
        marks=pytest.mark.xfail(
            reason="variable-length homogeneous tuple not supported"
        ),
    ),
```
- callables taking/returning custom pydantic models
```
    return (
        PM,
        val,
        name,
        {name: PM},
        {
            "stub generator produces incompatible method signatures for BaseModel subclasses"
        },
    )
``` 
Addresses #553

### DIFF
--- a/tests/test_handlers_llm_encoding.py
+++ b/tests/test_handlers_llm_encoding.py
@@ -1,12 +1,21 @@
 import builtins
+import dataclasses
+import random
 import typing
-from collections.abc import Callable
+from collections.abc import (
+    Callable,
+    Mapping,
+    MutableMapping,
+    MutableSequence,
+    Sequence,
+    Set,
+)
 from dataclasses import asdict, dataclass
 from typing import Annotated, Any, NamedTuple, TypedDict
 
 import pydantic
 import pytest
-from PIL import Image
+from PIL import Image, ImageChops
 from RestrictedPython import RestrictingNodeTransformer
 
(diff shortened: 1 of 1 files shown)


tests/test_handlers_llm_encoding.py
```diff
@@ -22,751 +31,347 @@
...
-        Encodable.define(Term)
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
 
 
-def test_type_to_encodable_type_operation():
-    with pytest.raises(TypeError):
-        Encodable.define(Operation)
+def _images_equal(a: Image.Image, b: Image.Image) -> bool:
```
eb8680: See #548 for how this is handled there. `Image.Image` implements `__eq__`, we should not be implementing our own.

PR #584, opened 2026-02-23
Replace `Encodable` implementations with Pydantic

eb8680:
This PR is another work-in-progress attempt at systematically fixing the myriad issues with `Encodable`. It preserves the shell of the `Encodable` API but breaks up the concrete `Encodable` implementations into custom validation, serialization and schema generation overrides within Pydantic's existing APIs that are injected by rewriting nested types, leaving only a single concrete `Encodable` that is a trivial wrapper around a `pydantic.TypeAdapter`.

There are still many kinks to work out but this seems like a promising alternative to the status quo in `master` and #548.


eb8680:
Subsumed by #619

Issue #569, opened 2026-02-15
Consider modifying `evaluate` to use object identity for DAG-based traversal


eb8680:
Hash-consing based on `__hash__` would be overly aggressive and unsound, but it should be reasonable to say that `evaluate` only descends into each host object once within a single traversal as determined by `id`.

eb8680:
This was resolved by #745

PR #594, opened 2026-02-24
Switch to DAG traversal of nested data structures in `evaluate`

eb8680:
Addresses #569


effectful/ops/semantics.py
```diff
@@ -147,12 +147,19 @@ def evaluate[T](
...
-    if intp is not None:
-        return interpreter(intp)(evaluate)(expr)
-
-    return __dispatch(type(expr))(expr)
+    from effectful.internals.runtime import get_runtime, interpreter
+
+    with interpreter(intp if intp is not None else get_runtime().interpretation):
+        cache = get_runtime().cache
+        assert cache is not None, "Cache should be initialized by interpreter"
+        key = id(expr)
```
jfeser, line 155: This is unsafe if terms are both allocated and deallocated during evaluation (id reuse). We should use a weak key dictionary instead.

jfeser (changes requested):

effectful/ops/semantics.py
eb8680: I think it's actually safe to use `id`s here because the keys are subterms in `expr`, so as long as `expr` is live its subterms are live. In other words, `id` should remain stable for the duration of a single `evaluate` call.

jfeser: It looks like the cache is reused in nested calls to `evaluate`, so a handler could create and evaluate multiple short-lived terms that collide in the cache.

eb8680: `cache[key] = (expr, result)` below so the result can never be deallocated within the outermost `evaluate` call if the cache is reused internally to that call

eb8680: I'm not sure how to use `weakref.WeakKeyDictionary` here. I guess we could make the cache a `WeakValueDictionary` whose values are `result`s?

jfeser: I think if all the terms in the cache remain alive until the cache is deallocated, then it should be safe. I'd add a comment to this effect.

It doesn't look like there's a straightforward use of `WeakKeyDictionary` that can deal with the need for a wrapper object to give the right equality/hash semantics.

jfeser (approved):

PR #604, opened 2026-03-10
Provides `PersistentAgent` class for Agent classes with persistent state (#603)

### LINKED ISSUE #603: More support for multi-agentic systems in `effectful`
Having been building more multi-agentic systems in Ceremony using effectful and recently coming across this blog post in HN:

https://rezzed.ai/blog/architecture-review-multi-agent-system

The notes from this blog post seem to match up closely with my experience from Ceremony, and seem like really useful features that would likely be of use across several projects.

At the very least, it would be particularly helpful to have a PersistentAgent subclass that persists state and can be resumed across sessions and also handles compaction of the state.

kiranandcode:
This PR provides a subclass to the effectful API which allows users to create `PersistentAgent`s who store their state to file and can be restored and resumed on crash. These agents also compact and restore their context. Closes #603 

```python
class ArchitectAgent(PersistentAgent):
    """You are a software architect. Given a project specification, you break
    it into individual module implementation tasks. Each task should specify
    the module filename, its public API, and what tests to write.
    Be concrete and specific — the coder will follow your spec exactly.
    """

    def __init__(self, persist_dir: Path):
        super().__init__(persist_dir=persist_dir, agent_id="architect")
        self._output_dir = OUTPUT_DIR

    @Tool.define
    def read_existing_files(self) -> str:
        """List files already written to the output directory."""
        if not self._output_dir.exists():
            return "No files yet."
        files = sorted(self._output_dir.rglob("*.py"))
        if not files:
            return "No Python files yet."
        return "\n".join(str(f.relative_to(self._output_dir)) for f in files)

    @Template.define
    def plan_modules(self, project_spec: str) -> PlanResult:
        """Given this project specification, output a plan with a "modules" list.
        Each module spec has: module_path, description, public_api, test_path.

        Use `read_existing_files` to check what's already been written
        and skip those.

        Project spec:
        {project_spec}"""
        raise NotHandled
```

See `docs/source/multi_agent_example.py` for an example of a non-trivial multi-agent workflow involving three agents doing a larger library task. This code is fairly robust, involves several agents, and can be Ctrl-C'd and restarted on demand.


kiranandcode:
@eb8680 , Hmm, yeah, that makes sense. I opted for class based state first because that makes tracking histories a little easier as a first pass than using a global handler and then re-associating calls to specific agents, but a global handler might be cleaner.

Will refactor.

docs/source/multi_agent_example.py
```diff
@@ -0,0 +1,537 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Worker loops
+# ---------------------------------------------------------------------------
+
+_shutdown = threading.Event()
+
+
+def architect_worker(queue: TaskQueue) -> None:
```
eb8680: Similar to design patterns we've seen in other examples (#547) and in building Pact, `architect_worker`, `coder_worker` and `reviewer_worker` seem like they ought to be obtainable mechanically as endpoint projections of a simpler global choreography, something like:
```py
def build_codebase(
  project_spec: ...
  architect: Architect,
  coder: Coder,
  reviewer: Reviewer,
):
  plan = architect.plan_modules(project_spec)
  code = coder.implement_module(plan)
  while True:
    verdict, feedback = reviewer.review_code(code, plan)
    if verdict == "NEEDS_FIXES":
      code = coder.implement_module(feedback)
    else:
      return code
```
These functions also mix application logic with the interfaces and implementations of several different side effects (`PersistentAgent`, `Template.__apply__`, `completion`, `LiteLLMProvider`, `RetryLLMHandler`, `sleep`, `claim`/`submit`/`complete`, `release_stale_claims`, `fail`, `_shutdown`, `log`/`exception`, `try`/`except`). We should be able to factor them into operations and handlers to reflect this structure. 

In addition to clarifying the coding application logic, this naturally leads to side effect interfaces and implementations that are modular, extensible and reusable across many contexts, not just this particular example.

eb8680:
Folded into #694

PR #612, opened 2026-03-13
Enable prompt caching for agent calls (closes #611)

### LINKED ISSUE #611: Effectful/litellm does not enable prompt_caching for anthropic
The Anthropic API has a mechanism for prompt-caching, where successive calls to llms with the same prefix will reuse the previous state of the LLM over the prefix instead of running from scratch. This is fundamentally necessary to enable long-running agentic interactions as otherwise as the agents persist for longer, the cost of each of their methods will increase linearly:

Image

Minimal reproducer:
```python

import collections
from effectful.handlers.llm.completions import (
    LiteLLMProvider, _make_message, completion, call_system, call_user,
)
from effectful.handlers.llm.template import Template
from effectful.ops.syntax import ObjectInterpretation, implements

captured = []

class Interceptor(ObjectInterpretation):
    @implements(completion)
    def _completion(self, *args, **kwargs):
        captured.append(kwargs.get("messages") or args[1])
        raise StopIteration

@Template.define
def ask(question: str) -> str:
    """You are a helpful assistant. Answer concisely.\n\n{question}"""

provider = LiteLLMProvider(model="anthropic/claude-sonnet-4-20250514")
try:
    with handler(provider):
        with handler(Interceptor()):
            ask("What is 2+2?")
except StopIteration:
    pass

def has_cache_control(msg):
    c = msg.get("content", "")
    if isinstance(c, list):
        return any("cache_control" in b for b in c if isinstance(b, dict))
    return "cache_control" in msg

print("=== Effectful Prompt Caching Report ===\n")
print(f"Messages sent to litellm.completion(): {len(captured[0])}")
for i, msg in enumerate(captured[0]):
    cached = has_cache_control(msg)
    print(f"  [{msg['role']:>9}] cache_control={cached}")
```

Outputs:
```python
=== Effectful Prompt Caching Report ===

Messages sent to litellm.completion(): 2
  [   system] cache_control=False
  [     user] cache_control=False
```

Relevant docs:

- Prompt Caching for Anthropic: https://platform.claude.com/docs/en/build-with-claude/prompt-caching

> Prompt caching optimizes your API usage by allowing resuming from specific prefixes in your prompts. This significantly reduces processing time and costs for repetitive tasks or prompts with consistent elements.
>
> Prompt caching stores KV cache representations and cryptographic hashes of cached content, but does not store the raw text of prompts or responses. This may be suitable for customers who require [ZDR-type data retention](https://platform.claude.com/docs/en/build-with-claude/zero-data-retention) commitments. See [cache lifetime](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#what-is-the-cache-lifetime) for details.
> 
> There are two ways to enable prompt caching:
> 
>    - [Automatic caching](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching): Add a single cache_control field at the top level of your request. The system automatically applies the cache breakpoint to the last cacheable block and moves it forward as conversations grow. Best for multi-turn conversations where the growing message history should be cached automatically.
> - [Explicit cache breakpoints](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#explicit-cache-breakpoints): Place cache_control directly on individual content blocks for fine-grained control over exactly what gets cached.

- OpenAI enables this by default: https://developers.openai.com/api/docs/guides/prompt-caching/

> Caching is enabled automatically for prompts that are 1024 tokens or longer. When you make an API request, the following steps occur:
>
> Cache Routing:
>
>    Requests are routed to a machine based on a hash of the initial prefix of the prompt. The hash typically uses the first 256 tokens, though the exact length varies depending on the model.
>    If you provide the [prompt_cache_key](https://developers.openai.com/api/docs/api-reference/responses/create#responses-create-prompt_cache_key) parameter, it is combined with the prefix hash, allowing you to influence routing and improve cache hit rates. This is especially beneficial when many requests share long, common prefixes.
>    If requests for the same prefix and prompt_cache_key combination exceed a certain rate (approximately 15 requests per minute), some may overflow and get routed to additional machines, reducing cache effectiveness.
>
>    Cache Lookup: The system checks if the initial portion (prefix) of your prompt exists in the cache on the selected machine.
>    Cache Hit: If a matching prefix is found, the system uses the cached result. This significantly decreases latency and reduces costs.
>    Cache Miss: If no matching prefix is found, the system processes your full prompt, caching the prefix afterward on that machine for future requests.

kiranandcode:
Updates `completion.py` such that system messages (call_system): content is now a list with cache_control: {"type": "ephemeral"} — cached across all turns.

Agent user messages (LiteLLMProvider._call): _add_cache_control_to_history() annotates the last user/tool message with cache_control before each call_assistant round — only for templates with __history__ (i.e. Agent subclasses)
Non-agent template calls are unaffected.

OpenAI calls are unaffected (OpenAI enables caching by default), `litellm` strips cache_control from OpenAI requests automatically.

Cost impact: Anthropic charges 25% more for cache writes but cached reads cost 90% less.


eb8680 (commented):
What happens if we add `cache_control` to each message at the time it is constructed? Is that incorrect vs the final-message-only behavior here?

effectful/handlers/llm/completions.py
```diff
@@ -159,6 +159,37 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
...
+    Mutates the history OrderedDict in place.
+    """
+    if not history:
+        return
+    for key in history:
+        msg = history[key]
+        if msg["role"] not in ("user", "tool", "assistant"):
+            continue
+        content = msg.get("content")
+        if isinstance(content, list) and content:
```
eb8680: Ah, I see this applies to all `Message`s. Would it be easier to have this live in `_make_message`, our `Message` constructor? Or maybe in `completions`?

kiranandcode: Ah, no I guess the key difference is that we only apply this if the template has a __history__ so we don't cache every template, only agent ones.

eb8680 (approved):

PR #613, opened 2026-03-13
Support typing.Self

### LINKED ISSUE #602: Type inference should dynamically resolve `typing.Self`
We'd like to be able to use `typing.Self` to annotate return types of method `Operation`s. From [the `typing` standard library documentation](https://docs.python.org/3/library/typing.html#typing.Self) and the [official `typing` specification](https://typing.python.org/en/latest/spec/generics.html#self), `Self` is equivalent to a type variable in the following sense, so it should be straightforward to fold into our existing machinery for handling type variables in `internals.unification`:
```python
class Foo:
  def return_self[Self: "Foo"](self: Self) -> Self:
    ...
    return self
```

eb8680:
Addresses #602


datvo06:
Have a question previously raised in #667 but I think worth checking again here. When we unify a type var with an inherited class, we are widening to the ancestor:
```
import typing                                                                                                                                                                                                                                                       
from effectful.internals.unification import unify                                                                                                                                                                                                                   
  
T = typing.TypeVar("T")                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                    
class Foo: pass
class Bar(Foo): pass

assert unify(T, Bar) == {T: Bar}      # FAILS: gets {T: Foo}          
```
Is it intended? I think by standard the principal type would be `Bar` instead of `Foo`?

eb8680:
> Is it intended? I think by standard the principal type would be Bar instead of Foo?

`unify` works up to the normal-form equivalence relation defined by `canonicalize`, which may attempt to widen types it does not know how to work with directly:
```py
assert unify(T, Bar) == unify(canonicalize(T), canonicalize(Bar)) == {T: canonicalize(Bar)}
```

PR #615, opened 2026-03-17
Adding anthropic agent provider

### LINKED ISSUE #614: support AgenticSDK as an LLM provider
Claude code token usage is separate from api key usage, we would ideally like to use claude code instead of api keys to maximise the tokens we have.

datvo06:
[DRAFT - Still need checking] closes #614

### DIFF
--- a/effectful/handlers/llm/anthropic.py
+++ b/effectful/handlers/llm/anthropic.py
@@ -2,7 +2,6 @@ import collections
 import collections.abc
 import inspect
 import json
-import textwrap
 import typing
 import uuid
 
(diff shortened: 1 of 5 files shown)


datvo06:
Still need a bunch of things to make sure it works:
1. Seems like Anthropic has different Tool call format than OpenAI. Maybe Encodable needs revisit (will check more carefully)
2. Need to setup CI to ensure integration with claude code pro/max work.

jfeser:
1. We should check that this is actually allowed by the terms of service for claude code.
2. There seems to be some LiteLLM support for oauth authentication, with the goal being to support this use case: https://github.com/BerriAI/litellm/issues/13380.

PR #619, opened 2026-03-23
Fix tuple schema compatibility with OpenAI structured output API

datvo06:
Addresses the prefixItems rejection from OpenAI for fixed-length tuple types (newly rediscovered in #617) by porting the Annotated-based approach.

Changes:
  - Add `_safe_tuple_type`: rewrites tuple[T1, T2, ...] into `Annotated[...,  PlainValidator, PlainSerializer, WithJsonSchema]` so schema, validation, and serialization are handled in one place via Pydantic's extension API
  - Add `_rewrite_tuple_annotations`: applies `_safe_tuple_type` recursively for
  dataclasses containing tuple fields
  - `TupleEncodable.encode/deserialize` now use `TupleItems` model instances consistently;
   shared field access via `_extract_items`
  - `SequenceEncodable.encode/deserialize` return lists to preserve encode idempotency
  - `_inline_refs`: inlines $ref pointers in WithJsonSchema values (workaround for
  pydantic/pydantic#12145)

`_safe_tuple_type` is a self-contained port of `_pydantic_type_tuple` from #584, scoped to the current Encodable system. `_rewrite_tuple_annotations` handles the dataclass-with-tuple-field case until #584 lands.


eb8680 (commented):
I am a bit confused by the purpose of this PR. There are no new tests that were failing without and passing with the behavior changes, and there are existing tests xfailing with the changes that previously passed. Did we not fix the OpenAI API tuple encoding already? What is new here?

It also seems to have introduced a fragile new invariant that `Encodable.encode` must never return tuples or values that contain tuples that are not fields of some enclosing `BaseModel`, but that's not documented anywhere and can't be checked statically by a type checker.

datvo06:
Update: Fixes #626, #631. Supersedes #584.                              
- Summary                                                                                                                                                                                                                                                                                                                                                                                                                                                       
  - Replace the hand-rolled Encodable ABC and its 10+ subclasses with TypeToPydanticType, a recursive type rewriter that produces Annotated wrappers with Pydantic validators/serializers                                                                                                                                                                                                                                                                               
  - Add TypeEvaluator ABC to unification.py - a generic recursive type-expression walker using singledispatchmethod
  - Migrate all Encodable.define(T).encode/decode call sites to TypeAdapter(Encodable[T]).dump_python/validate_python                                                                                                                                                                                                                                                                                                                                                   
  - Extract _collect_tools() and to_content_blocks() helpers in completions.py                                                                                                                                                                                                                                                                                                                                                                                          

Why                                                                                                                                                                                                                                                                                                                                                                                                                                                                   
- The old system had each Encodable subclass reimplementing Pydantic's validation/serialization by hand. It didn't recurse into field types, so dataclass fields like tuple[int, int] (#626) or Callable[[int], int] (#631) produced invalid JSON schemas. The new approach fixes both bugs automatically because the recursion rewrites field annotations before Pydantic sees them.

datvo06:
One caveat is composite type field rewriting is not very clean: To handle field annotation inside dataclasses and basemodels, those fields contain types that need custom encoding (e.g., Callable, tuple, Image). It builds a proxy Pydantic model with the rewritten annotations, then wraps the original type in Annotated[T, PlainValidator, PlainSerializer, WithJsonSchema] so roundtrip preserves the original type.                                                                                             
                                                                                                                                                                                                                                                                                                                                                                                                                                                                        
- Alternatives considered:                                                                                                                                                                                                                                                                                                                                                                                                                                
  - Option A (no rewriting): Don't handle fields at all. Users annotate special fields themselves with Encodable[...]. Zero custom code but doesn't fix #631 for plain dataclasses.                                                                                                                                                                                                                                                                                     
  - Option B (subclass): Create a pydantic dataclass subclass with rewritten `__annotations__`. ~10 lines, but `type(decoded) != type(original)` - breaks `__eq__` for standard dataclasses.                                                                                                                                                                                                                                                                                
  - Option C (current): Proxy model + Annotated validators. ~40 lines per handler. Preserves exact type identity. Uses standard Pydantic extension points (`PlainValidator, PlainSerializer, WithJsonSchema`).

tests/test_handlers_llm_encoding.py
```diff
@@ -815,11 +952,17 @@ def _fn():
     _fn.__annotations__ = {"return": ty}
 
     tool: Tool[..., Any] = Tool.define(_fn)
+    enc: pydantic.TypeAdapter[Any] = pydantic.TypeAdapter(
+        Encodable[type(tool)]  # type: ignore[misc]
+    )
+    tool_spec = _strict_json_schema(
```
eb8680: This should happen internally when applying `Encodable` to `Tool`. Tests and user code should never have to invoke `_strict_json_schema` directly.

tests/test_handlers_llm_encoding.py
```diff
@@ -790,18 +921,24 @@ def _fn(value):
     _fn.__annotations__ = {"value": ty, "return": None}
 
     tool: Tool[..., Any] = Tool.define(_fn)
+    enc: pydantic.TypeAdapter[Any] = pydantic.TypeAdapter(
+        Encodable[type(tool)]  # type: ignore[misc]
+    )
+    tool_spec = _strict_json_schema(
```
eb8680: This should happen internally when applying `Encodable` to `Tool`. Tests and user code should never have to invoke `_strict_json_schema` directly.

tests/test_handlers_llm_encoding.py
```diff
@@ -706,77 +824,90 @@ def __call__(self):
...
             out.append(c)
-            continue
-        out.append(pytest.param(*c.values, id=case_id, marks=marks))
     return out
 
 
-PROVIDER_CASES = _cases_with_provider_xfails(ROUNDTRIP_CASES)
-
+# response_model: image types can't roundtrip (LLM returns URLs, not data URIs),
+# and Tool/DecodedToolCall schemas are incompatible with OpenAI strict mode.
```
eb8680: This comment about `Tool` seems like a bug, it should be fixed rather than `xfail`ed.

tests/test_handlers_llm_encoding.py
```diff
@@ -706,77 +824,90 @@ def __call__(self):
...
-def _encode_tool_spec(tool: Tool[..., Any]) -> dict[str, Any]:
-    tool_ty: type[Any] = type(tool)
-    tool_enc: Encodable[Any, Any] = Encodable.define(tool_ty)
-    tool_spec_obj = tool_enc.encode(tool)
-    if isinstance(tool_spec_obj, Mapping):
-        return dict(tool_spec_obj)
-    elif hasattr(tool_spec_obj, "model_dump"):
-        return dict(tool_spec_obj.model_dump())
-    raise TypeError(f"Unexpected encoded tool spec type: {type(tool_spec_obj)}")
+# tool-as-param: only Tool/DecodedToolCall schemas fail (nested function spec).
```
eb8680: This comment about `Tool` seems like a bug, it should be fixed rather than `xfail`ed.

tests/test_handlers_llm_encoding.py
```diff
@@ -706,77 +824,90 @@ def __call__(self):
...
+    schema: dict[str, Any] = _strict_json_schema(
+        {
+            "type": "object",
+            "properties": {"value": inner_schema},
+            "required": ["value"],
+        }
+    )
+    response = litellm.completion(
+        model=EFFECTFUL_LLM_MODEL,
+        response_format={
```
eb8680: This should probably be a Pydantic model rather than a raw JSON schema.

effectful/handlers/llm/encoding.py
```diff
@@ -62,374 +103,309 @@ class DecodedToolCall[T]:
...
+        pydantic.PlainValidator(_validate),
+        pydantic.PlainSerializer(_serialize),
+        pydantic.WithJsonSchema(
+            _strict_json_schema(_inline_refs(proxy_adapter.json_schema()))
+        ),
+    ]
+
+
+@TypeToPydanticType.register(pydantic.BaseModel)
+def _pydantic_type_basemodel(ty: type[pydantic.BaseModel]) -> Any:
```
eb8680: I do not think we should be doing this transformation for `pydantic.BaseModel`s, which are in charge of their own serialization/validation semantics and for which this is highly unlikely to be correct in general. Injecting `Encodable` into a `BaseModel` should be done manually by a user writing appropriate annotations.

effectful/handlers/llm/encoding.py
```diff
@@ -62,374 +103,309 @@ class DecodedToolCall[T]:
...
+        construct=ty,
+        extract=lambda v, ns: {n: getattr(v, n) for n in ns},
+    )
 
-    def deserialize(self, serialized_value: str) -> str:
-        return serialized_value
 
+@TypeToPydanticType.register(object)
+def _pydantic_type_base(ty: type) -> Any:
+    if dataclasses.is_dataclass(ty) and isinstance(ty, type):
```
eb8680: I am skeptical of the correctness of this transformation for general polymorphic `dataclass`es, and it has almost no test coverage. I would strongly prefer to remove it and require users to add `Encodable` annotations manually unless/until we come up with a more convincing specification for this kind of automated behavior.

effectful/handlers/llm/encoding.py
```diff
@@ -62,374 +103,309 @@ class DecodedToolCall[T]:
...
+@TypeToPydanticType.register(str)
+def _pydantic_type_str[T](ty: type[T]) -> type[T]:
+    return ty
 
-    def encode(self, value: str) -> str:
-        return value
 
-    def decode(self, encoded_value: str) -> str:
-        return encoded_value
+def _rewrite_fields(
```
eb8680: I don't think this is correct in general, especially not shared across different record type families, and I suspect it would be very difficult to cover all the edge cases, especially once polymorphism is included. I would strongly prefer to remove it from this PR in order to keep the changes tractable and sound.

effectful/handlers/llm/encoding.py
```diff
@@ -62,374 +103,309 @@ class DecodedToolCall[T]:
...
-        return tuple(
-            getattr(model, f"item_{i}") for i in range(len(self.element_encoders))
-        )
+    return _resolve(schema)
 
 
-@dataclass
-class NamedTupleEncodable[T](TupleEncodable[T]):
-    """Tuple encodable that reconstructs the original NamedTuple type on decode."""
+def _strict_json_schema(schema: dict) -> dict:
```
eb8680: There seem to be two versions of this function. I suspect this transformation is not actually necessary to begin with given that it has not been a problem before - perhaps `litellm` or `pydantic` enforces it somehow.

effectful/handlers/llm/completions.py
```diff
@@ -172,10 +221,70 @@ def completion(*args, **kwargs) -> typing.Any:
     return litellm.completion(*args, **kwargs)
 
 
+class _BoxedResponse[T](pydantic.BaseModel):
+    value: T
+
+
+def _strict_json_schema(schema: dict) -> dict:
```
eb8680: There seem to be two versions of this function (this one and one in `encoding.py`). I also suspect it is not necessary or at least is already available in some `litellm` utility, otherwise nothing would have worked before with the OpenAI API.

effectful/handlers/llm/completions.py
```diff
@@ -172,10 +221,70 @@ def completion(*args, **kwargs) -> typing.Any:
...
+    """Adapt a JSON schema for OpenAI's structured output strict mode.
+
+    Applies all transformations required by the OpenAI strict-mode API:
+
+    * Inlines ``$ref``/``$defs`` (OpenAI does not support JSON Schema references).
```
eb8680: We have never had issues with this before, either, which strengthens my suspicion that `litellm` or `pydantic` are doing this somewhere already. The `_inline_refs` I originally implemented in #584 was a workaround for a Pydantic bug, not a necessary invariant for OpenAI API compatibility.

eb8680 (changes requested):
I don't think we should be attempting to transform arbitrary record types. One motivation for the new `Encodable` interface in #584 was to avoid the need for this, at least to start.

There is also a bunch of schema-munging happening that is likely not necessary.

effectful/handlers/llm/encoding.py
datvo06: I found one provided by OpenAI and replaced it!

effectful/handlers/llm/encoding.py
datvo06: Got it. I removed it and added dataclass encodable in `PairManual` (dataclass with tuple) in the test.

effectful/handlers/llm/encoding.py
datvo06: I updated it to only handle base type (return the type) instead of rewriting record types.

tests/test_handlers_llm_encoding.py
datvo06: On master, Tool/DecodedToolCall were xfailed for response_format integration tests. We maintain the same xfails. We still support encoding/deserializing them for tool calling.

Tool as `response_format` fails because OpenAI rejects `ChatCompletionToolParam`'s schema (bare "type": "object" without additionalProperties: false). 
Fixing it is non-trivial and gory because we'd have to write a recursive schema traversal and add that additionalProperties annotation.

I tried finding tools from Litellm, OpenAI, or Pydantic, but none of them provide exact tools for this.

datvo06:
I did more cleaning up. Now the dataclass encodable is defined by user, it will be fine as long as user have dataclass fields being encodable:
```python
@dataclass
class _PairManual:
    values: Encodable[tuple[int, str]]
    count: int
```

But there is still one test marked xfail (both here and on master), which is the round-trip Tool and DecodedToolCall.

On master, Tool/DecodedToolCall were xfailed for response_format integration tests. We maintain the same xfails. We still support encoding/deserializing them for tool calling.

Tool as `response_format` fails because OpenAI rejects `ChatCompletionToolParam`'s schema (bare "type": "object" without additionalProperties: false). 
Fixing it is non-trivial and gory because we'd have to write a recursive schema traversal and add that additionalProperties annotation.

I tried finding tools from Litellm, OpenAI, or Pydantic, but none of them provide exact tools for this.

eb8680 (approved):
Still not sure I understand the problems with `Tool`/`DecodedToolCall`, but I think they can be fixed later. Otherwise I think this is good to go. Thanks for pushing it through!

PR #634, opened 2026-04-10
Remove ignore: type that's triggering an error

yiyunliu:
Just a dummy PR to see what error message the CI generates. Running `act` locally isn't feasible because I need the specific configuration cached by the CI, which sadly can't be accessed directly.

### DIFF
--- a/effectful/handlers/llm/evaluation.py
+++ b/effectful/handlers/llm/evaluation.py
@@ -392,7 +392,7 @@ def signature_to_ast(name: str, sig: inspect.Signature) -> ast.FunctionDef:
         except TypeError:
             returns = type_to_ast(typing.Any)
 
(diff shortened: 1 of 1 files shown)


yiyunliu:
The issue has been fixed. Mypy should now be happy regardless of the Python version. I opened a separate issue #635 regarding the CI using old caches, which hopefully will make bugs like this easier to fix in the future.

The notebook and llm checks failed but according to the log the failures were caused by missing openai keys. (I think that has something to do with me not being a contributor to effectful?)

eb8680:
> The notebook and llm checks failed but according to the log the failures were caused by missing openai keys. (I think that has something to do with me not being a contributor to effectful?)

You should generally be pushing to branches in `BasisResearch/effectful`, not a fork of the repository.

eb8680 (approved):

PR #637, opened 2026-04-13
calculate cache key from pyproject.toml

### LINKED ISSUE #635: CI cache key refers to non-existent uv.lock file
The current CI configuration (e.g. [lint.yml](https://github.com/BasisResearch/effectful/blob/master/.github/workflows/lint.yml)) constructs the cache key by hashing `uv.lock` file, which isn't in version control.

The consequence is that the cache key is a constant (see [the CI caches page](https://github.com/BasisResearch/effectful/actions/caches)) and never gets invalidated no matter how the project worktree is modified . Not having the CI package versions visible makes issues like #630 hard to debug, when there's a discrepancy between the CI environment and a fresh local environment.

I can create a PR and commit a `uv.lock` file.

yiyunliu:
uv.lock doesn't exist and we want the cache to properly refresh when we change pyproject.toml, which is the de facto source for supported package versions.

Fixes #635

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,15 +15,9 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v5
+        uses: astral-sh/setup-uv@v7
         with:
           enable-cache: true
-
-      - name: Cache virtual environment
-        uses: actions/cache@v4
-        with:
-          path: .venv
-          key: venv-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
 
       - name: Install dependencies
         run: |
@@ -31,6 +25,3 @@ jobs:
 
       - name: Lint
         run: uv run ./scripts/lint.sh
-
-      - name: Minimize uv cache
-        run: uv cache prune --ci
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -31,3 +31,6 @@ jobs:
 
       - name: Lint
         run: uv run ./scripts/lint.sh
+
+      - name: Minimize uv cache
+        run: uv cache prune --ci
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,7 +23,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: .venv
-          key: venv-${{ runner.os }}-${{ hashFiles('uv.lock') }}
+          key: venv-${{ runner.os }}-${{ hashFiles('pyproject.toml') }}
 
       - name: Install dependencies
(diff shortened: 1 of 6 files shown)


yiyunliu:
Looking at https://github.com/BasisResearch/effectful/actions/caches, the hashkey was indeed calculated properly.

kiranandcode (commented):
LGTM; will await review from @jfeser for a second look from someone more familiar with the CI setup?

yiyunliu:
Another cache-related change that may make sense to also add to the PR: `test.yml` now uses a matrix with two Python version numbers but `lint.yml` doesn't. This ends up creating 3 caches, two with version numbers `3.12` and `3.13` in their hash keys and one from `lint.yml` without python versions attached but likely uses either `3.12` or `3.13`. We can save lots of cache space just by appending python version numbers to all workflows. Right now we are quite close to the cache size limit.

jfeser:
I actually think we'd be better off disabling the cache. `uv` is very fast and we don't build any wheels. Compare https://github.com/BasisResearch/effectful/actions/runs/23057444832/job/66974659394, which takes 1min to restore the cache with https://github.com/BasisResearch/effectful/actions/runs/24349662751/job/71100186595 which takes 22s to build the venv from scratch.

yiyunliu:
So I did some quick googling and found [this](https://docs.astral.sh/uv/concepts/cache/#caching-in-continuous-integration) from uv's documentation.

In short, `uv` provides the `uv cache prune --ci` command which removes pre-built wheels and keeps only ones that were built from source. We could just add that to the CI to make it more future-proof, although if we really aren't building any from source then the cache would just be mostly empty.

eb8680:
Caching definitely sped things up in #595, it was taking substantially longer to download and install everything before that. I'm not sure why it looks differently or what's changed in the two logs you're showing but the lint and doc build steps in particular were taking 2 minutes or more to get through all the setup versus just a few seconds on the actual work.

yiyunliu:
I'll push a commit with `uv cache prune --ci` and see how it looks. If I understand the documentation correctly, it should give us the best of both worlds (caching everything vs. no cache).

yiyunliu:
So I realize `uv` already knows how to cache things properly, including using the project file (`pyproject.toml`) and python version to automatically generate shared cache. I ended up removing the custom `venv` cache and just let `uv` handle caching on its own.

By default, the `uv` setup step with cache enabled already prunes the pre-built wheels. If you look at the generated cache from the latest workflow triggered by the PR, you'll see that each cache take very little space. That agrees with @jfeser suspected earlier: if no wheel was built from source then `uv` doesn't need to store anything.

I still think it would make more sense to have the PR merged and just let `uv` do the right thing.

jfeser (approved):

Issue #493, opened 2026-01-19
Consider offloading observability to LiteLLM


eb8680:
Currently `handlers.llm` contains a somewhat half-baked implementation of API call logging in `handlers.llm.providers`. We should switch to [LiteLLM's built-in observability functionality](https://docs.litellm.ai/docs/observability/callbacks) rather than trying to implement more of it ourselves (#391), unless it becomes clear that we need some additional behavior.

Issue #391, opened 2025-11-07
Support for call tracing API's like `weave`


kiranandcode:
The following code works:
```python
provider = OpenAIAPIProvider(client = OpenAI())
weave.init('experiments')

@Template.define
@weave.op()
def limerick(topic: str) -> str:
    """Construct a limerick about {topic}."""
    raise NotHandled
    
with handler(provider):
    print(limerick("abstract interpretation of floats"))
```
with the minor issue that the name of the function is not present in the logs:
<img width="979" height="476" alt="Image" src="https://github.com/user-attachments/assets/fe926ce5-41bf-449a-ade3-79b04721c9d0" />

(swapping the order of the decorators results in an exception)

when building experiments using LLM's effectful API, libraries such as weave can be particularly important for being able to inspect results.

PR #638, opened 2026-04-14
add weave and langfuse observability support

yiyunliu:
This PR adds observability support (#605) through Langfuse and Weave. The implementation simply instruments template, tool, and other llm related operations to use the decorator provided by the respective framework. Langfuse requires a bit more effort because its native litellm support doesn't seem compatible with custom spans for effectful handlers. I had to write some code to populate cost and token information from the litellm response.

This PR is different from #632, which aims to provide more flexibility by hand-rolling some observability functionalities so they can be accessed by the programmer directly.

### DIFF
--- a/effectful/handlers/llm/langfuse.py
+++ b/effectful/handlers/llm/langfuse.py
@@ -1,0 +1,81 @@
+import functools
+import typing
+
+import litellm
+from langfuse import get_client, observe
+
+from effectful.handlers.llm import Template, Tool
+from effectful.handlers.llm.completions import (
+    call_assistant,
+    call_system,
+    call_user,
+    completion,
+)
+from effectful.ops.semantics import fwd
+from effectful.ops.syntax import ObjectInterpretation, implements
+
+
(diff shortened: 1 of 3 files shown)


kiranandcode:
Possibly relevant: https://github.com/BasisResearch/effectful/issues/391

maybe things have changed since then.

yiyunliu:
#391 was closed in favor of #493, which suggests that effectful should use litellm's logging backend. I did try litellm's `langfuse_otel` backend through the callback interface. The issue is that you only see the LLM API calls but you don't see any of the template or tool traces unless you instrument those functions yourself. The same happens with `weave` I think.

yiyunliu:
To provide more context, here's a screenshot of langfuse with litellm's native callback support after calling `limerick`:
llmlitenative

and a screenshot of langfuse with call traces produced by the PR:
instrumented

eb8680:
Closing in favor of the more correct langfuse API usage in #694

PR #641, opened 2026-04-21
Add an op for dataclass construction

### LINKED ISSUE #624: Term construction fails on dataclass arguments
The following test should pass:

```py
def test_defdata_dataclass() -> None:
    @Operation.define
    def f(x: int):
        raise NotHandled

    @dataclasses.dataclass
    class A:
        x: int

        def __init__(self, x: int):
            self.x = f(x)

    @Operation.define
    def g(a: A):
        raise NotHandled

    v = Operation.define(int)
    t = g(A(v()))
    assert isinstance(t.args[0].x, Term)
```

jfeser:
Closes #624


jfeser:
Subsumed by #719

PR #646, opened 2026-04-24
Unpin litellm dependency in pyproject.toml

eb8680:
Removed version constraint for `litellm` package that isn't doing anything useful anymore.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,7 @@ numpyro = [
   "jax<0.10"
 ]
 llm = [
-  "litellm<1.82.7",
+  "litellm",
   "tenacity",
   "mypy",
   "autoflake",


datvo06 (approved):

datvo06:
For context (if we'd ever stumbled upon the PR again). This is mostly pulled from Claude, but I reviewed the link and confirmed.

Removing the pin is safe now. Here's the timeline:                                                                                                                                                                                                                                                                                     
  - 2026-03-24 (~40 min window): attackers published malicious litellm 1.82.7 and 1.82.8 to PyPI. The .pth payload ran credential-stealing code on every Python startup. PyPI quarantined the package within hours.
  - Both 1.82.7 and 1.82.8 have been removed/yanked from PyPI.                                                                                                          
  - 2026-03-30: clean release 1.83.0 shipped through a hardened CI/CD pipeline. LiteLLM has audited every release between v1.78.0 and v1.82.6 and confirmed them clean. 
                                                                                                                                                                        
The existing pin in pyproject.toml was litellm<1.82.7 - so it only ever excluded the two yanked versions, not the clean 1.83.x line. That means:                                                                                                               
  1. The dangerous versions are gone from the index, so a normal resolve can't pick them up anyway.                                                                     
  2. The pin never blocked 1.83.0+, which is where the fix actually lives.
  3. Therefore the pin is doing zero defensive work today - removing it is a no-op from a security standpoint, matching the PR description.                             
                                                                                                                                                                        
  One small caveat worth knowing (not a blocker for merging #646): yanked versions can still be installed if someone types litellm==1.82.7 explicitly or already has the wheel in a local cache. Neither the old pin nor removing it affects that - both scenarios are out of scope for a version constraint. If you want belt-and-suspenders, you could add a lower-bound pin litellm>=1.83.0 instead of unpinning entirely, but that's optional; the yank + the new clean release are the real fix.               

  Sources:
  - [Security Update: Suspected Supply Chain Incident (LiteLLM blog, March 2026)](https://docs.litellm.ai/blog/security-update-march-2026)[https://docs.litellm.ai/blog/security-update-march-2026](vscode-file://vscode-app/Applications/Cursor.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
  - [litellm 1.82.8 Supply Chain Attack on PyPI (FutureSearch) ](https://futuresearch.ai/blog/litellm-pypi-supply-chain-attack/)[https://futuresearch.ai/blog/litellm-pypi-supply-chain-attack/](vscode-file://vscode-app/Applications/Cursor.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)                                                                                                          
  - GitHub issue BerriAI/litellm#24512

PR #650, opened 2026-04-27
Fix branch in `nested_type` for Mapping handler

kiranandcode:
For the handler for `Mapping`, branch should be `len(value) == 2`, (len(value) for a mapping is never 1).

### DIFF
--- a/effectful/internals/unification.py
+++ b/effectful/internals/unification.py
@@ -852,7 +852,7 @@ def _(value: collections.abc.Mapping):
 
     if len(value) == 0:
         return Box(type(value))
-    elif len(value) == 1:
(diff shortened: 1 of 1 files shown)


kiranandcode:
though overall this branch seems redundant (as evidenced by the fact that it was never being taken before)

eb8680:
> For the handler for Mapping, branch should be len(value) == 2, (len(value) for a mapping is never 1).

This is not correct - `len(value)` is the number of key-value pairs, e.g. `len({"a": 1}) == 1`.

kiranandcode:
oh, my bad, I must have misinterpreted the len output.

PR #651, opened 2026-04-27
updated nested type to work with typedicts (#583)

### LINKED ISSUE #583: Nested Type should work with TypedDict
Currently, we would expect `nested_type` to return a typeddict type annotation on a `TypedDict` instance. 
```python
from typing import TypedDict
from effectful.internals.unification import nested_type

class UserTD(TypedDict):
    name: str
    age: int

print("nested_type(UserTD(name='a', age=1)).value ->", nested_type(UserTD(name='a', age=1)).value)
```
Result:
```
nested_type(UserTD(name='a', age=1)).value -> <class 'dict'>
```

One difficulty is that the TypedDict instance at runtime is just a `dict`. Inferring types is possible, but we might not be able to recover the schema fully.

kiranandcode:
Closes #583

```python
from typing import TypedDict
from effectful.internals.unification import nested_type

class UserTD(TypedDict):
    name: str
    age: int

print("nested_type(UserTD(name='a', age=1)).value ->", nested_type(UserTD(name='a', age=1)).value)

```

Now returns:
```python
nested_type(UserTD(name='a', age=1)).value -> <class 'effectful.internals.unification.RuntimeTypeDict'>
```


eb8680:
I think for this to be useful it has to be accompanied by support for `TypedDict` in `unify`:
```py
class Datum[T](TypedDict):
    name: str
    value: T

...

assert unify(Datum[T], nested_type({"name": "a", "value": 1}).value) == {T: int}
```
Without that additional change, this is just going to break `typeof` for `Mapping`s, because values will get turned into `TypedDict`s that `unify` doesn't know how to handle.

kiranandcode:
@eb8680 updated to handle typedicts more extensively!

effectful/internals/unification.py
```diff
@@ -262,6 +262,9 @@ def unify(typ, subtyp, subs: Substitutions = {}) -> Substitutions:
     if isinstance(typ, inspect.Signature):
         return _unify_signature(typ, subtyp, subs)
 
+    if _is_typeddict_type(typ) and _is_typeddict_type(subtyp):
+        return _unify_typeddict(typ, subtyp, subs)
```
eb8680, line 269: Nothing else in `unify` should happen before the `canonicalize` calls below - `unify` is only sound up to the normal form defined by `canonicalize`. This needs to move down and `canonicalize` needs to support `TypedDict`.

tests/test_internals_unification.py
```diff
@@ -188,6 +188,16 @@ class GenericClass[T]:
...
 
 
+def test_canonicalize_typeddict_class():
+    """Canonicalizing a TypedDict class returns MutableMapping."""
+
+    class MyTD(typing.TypedDict):
+        name: str
+        age: int
+
+    assert canonicalize(MyTD) == collections.abc.MutableMapping
```
eb8680: I think this behavior also needs to change. If we're now supporting `TypedDict`s in `unify`, `canonicalize` shouldn't eliminate it, it should normalize the fields and return a new `TypedDict`.

tests/test_internals_unification.py
```diff
@@ -1769,3 +1831,14 @@ def test_binary_on_sequence_elements(f, seq, index1, index2):
         ),
         collections.abc.Mapping,
     )
+
+
+def test_unify_generic_typeddict_extracts_typevar():
```
eb8680, line 1876: This needs quite a few more test cases to account for the full behavior of `TypedDict`, e.g. its interaction in type inference with `Mapping` and its`total` keyword and `Required`/`NotRequired` special forms: https://docs.python.org/3/library/typing.html#typing.TypedDict

In order to preserve the overall soundness of `unify`, if you don't want to implement support for those behaviors to make the additional test cases pass, you should make sure they either trigger failures or are eliminated in `canonicalize`.

effectful/internals/unification.py
kiranandcode: Gotcha, yes, that makes sense.

eb8680 (commented):
I asked Claude to review against the official spec and it had many complaints:

> Review the correctness of these changes (soundness and completeness) with respect to the official Python typing spec for `TypedDict`: https://typing.python.org/en/latest/spec/typeddict.html 
> 
> Be sure to review the tests added in @tests/test_internals_unification.py  and the new logic in @effectful/internals/unification.py  . Remember to check for soundness violations in `unify` introduced by the new changes.

## Summary

The PR adds structural unification for TypedDicts plus runtime TypedDict inference in `nested_type`. The basic shape is reasonable, but **`_unify_typeddict` is unsound in several respects** vs. the typing spec, and the new tests bake those unsound behaviors in as expected.

## Soundness violations

The spec ([typeddict.html § Subtyping](https://typing.python.org/en/latest/spec/typeddict.html)) says, for a subtype B and supertype A:

> "If [an item] is required in A, it must also be required in B."
> "If it is mutable in A, it must also be mutable in B, and the item type in B must be equivalent to the item type in A."
> "[For non-required mutable items in A:] If B has [the] item, it must also be mutable and non-required, and its item type must be equivalent."

### 1. Required-ness asymmetry not enforced — `effectful/internals/unification.py:395-420`

The loop iterates `typ_hints` and only fails when a key in the supertype is *missing* from the subtype. It never fails when a `Required` key in `typ` is present-but-`NotRequired` in `subtyp`, nor when a `NotRequired` key in `typ` is `Required` in `subtyp`.

The new test `test_unify_typeddict_notrequired_present` (`tests/test_internals_unification.py:1888-1900`) hard-codes the unsound case:

```python
class Pattern(TypedDict):
    nickname: NotRequired[str]   # non-required, mutable in supertype
class Sub(TypedDict):
    nickname: str                # Required in subtype
unify(Pattern, Sub)   # passes, but spec rejects:
                      # "B [Sub] must also be mutable and non-required"
```

The same flaw is in the other direction — a `Required[name]` pattern with a `NotRequired` `name` in `subtyp` would also pass.

To honor the spec, `_unify_typeddict` needs to compare `__required_keys__`/`__optional_keys__` between `typ` and `subtyp` for each common key, not only check membership.

### 2. Field types unified covariantly, but the spec mandates *equivalence* (invariance) for mutable items — `effectful/internals/unification.py:418`

`unify(field_type, subtyp_hints[field], subs)` is the regular subtype-direction unification (covariant via the `_unify_generic` subclass branch). For mutable TypedDict fields the spec requires the value types to be *equivalent*. So:

```python
class Pattern(TypedDict):
    items: list[object]
class Sub(TypedDict):
    items: list[int]
unify(Pattern, Sub)   # currently succeeds (list is mutable, invariance required by spec)
```

slips through because `_unify_generic` accepts `list[int]` as a subtype of `list[object]` via the `issubclass(int, object)` walk.

A correct fix needs a separate "equivalence" check (bidirectional unify, or equality after canonicalization) for mutable fields, and only fall back to subtyping for `ReadOnly[...]` items.

### 3. `Mapping[K, V]` vs. TypedDict falls through to the subclass shortcut — `effectful/internals/unification.py:470-475`

Spec: *"A TypedDict type is a subtype of `Mapping[str, VT]` when all value types of the items in the TypedDict are subtypes of VT."*

But `_unify_generic` reaches this branch:

```python
elif (
    isinstance(typ, GenericAlias)
    and isinstance(subtyp, type)
    and issubclass(subtyp, typing.get_origin(typ))
):
    return subs  # implicit expansion to subtyp[Any]
```

`issubclass(MyTD, Mapping)` is True (TypedDict inherits `dict`), so e.g. `unify(Mapping[str, int], InfoTD_with_str_fields)` returns `{}` and never inspects field types. The new test `test_unify_mapping_empty_typeddict` (`tests/test_internals_unification.py:2031-2039`) only exercises the `Empty` case, which masks this; but combining the new TypedDict awareness with the existing fallthrough makes the unsoundness more reachable. Adding a TypedDict-aware case in `_unify_generic` (iterate field hints, unify each against `V`) would close it.

### 4. `ReadOnly` modifier ignored entirely

The spec uses `ReadOnly` to *legally* relax invariance to covariance and to allow non-required→required overrides. The implementation doesn't read `__readonly_keys__` (PEP 705) at all, so when proper invariance is added (issue 2) it will conservatively reject valid `ReadOnly` covariant subtypes. Today this is masked because invariance isn't enforced in the first place.

## Completeness / correctness gaps

### 5. Generic-via-inheritance TypedDicts aren't substituted — `effectful/internals/unification.py:382-392`

`_get_typeddict_hints` only substitutes when `typing.get_origin(typ) is not None`. For `class Derived(Base[int]): ...`, `typing.get_origin(Derived)` is `None`, so it falls into `else: typing.get_type_hints(typ)`. The test `test_unify_typeddict_inheritance_generic` (`tests/test_internals_unification.py:2011`) literally documents this:

```python
# T from Base is still in Derived's hints, so it gets unified with int
assert any(v == int for v in subs.values())
```

That test only passes by accident — the leftover `T` happens to land on `int` during unify. A peer with a non-generic `Sub` field, or a value type reused across fields with conflicting bindings, will give the wrong answer or spurious unification failures. Proper handling requires walking `__orig_bases__` and substituting through the generic base chain (or relying on `typing.get_type_hints` with the right form, which doesn't substitute today).

### 6. `nested_type` TypedDict synthesis is inconsistent — `effectful/internals/unification.py:919-946`

- `{"name": "Alice"}` (len=1, str→str) → `MutableMapping[str, str]`
- `{"name": "Alice", "age": 30}` (len=2, mixed values) → `TypedDict`
- `{"name": "Alice", "age": "Bob"}` (len=2, both str) → `MutableMapping[str, str]`

So a one-key dict of the same shape, or a multi-key dict whose values happen to share a type, never becomes a TypedDict. The trigger is "values are heterogeneous enough to produce a `UnionType`," which is a brittle proxy for "the user intended a TypedDict." Probably fine for the original `nested_type(UserTD(...))` use case (which always has heterogeneous fields), but worth flagging.

### 7. `canonicalize` rebuilds non-canonical TypedDicts as a fresh class named `"_Canonical"` — `effectful/internals/unification.py:575-589`

Every non-canonical TypedDict canonicalizes to a brand-new `typing.TypedDict("_Canonical", ...)`. Two unrelated user TypedDicts both turn into different classes that share the name `_Canonical`, which makes error messages ambiguous and causes `==` to fail across calls (since TypedDict equality is identity). Idempotency is preserved (the test verifies `is`), but the original TypedDict's `__name__` is lost, and repeated calls in hot paths leak class objects. Preserving `typ.__name__` (or returning the original when only inner-arg changes occur) would help.

## Things that are fine

- Width subtyping direction (iterating `typ_hints` only, allowing extra keys in `subtyp`) matches the spec.
- `__optional_keys__` is the right attribute to read for `total=False`, plain TypedDict, and `Required[...]` overrides.
- Inheritance for the non-generic case (`test_unify_typeddict_inheritance`, `tests/test_internals_unification.py:1980-1994`) works because `get_type_hints` walks the MRO.
- `canonicalize` for `NotRequired[...]` fields preserves the optional-ness via re-wrapping (`effectful/internals/unification.py:585-588`). The PR's own `test_canonicalize_typeddict_notrequired` (`tests/test_internals_unification.py:2042`) confirms this.
- The `ForwardRef` fix (`getattr(typ, "__forward_value__", None)`) is unrelated to TypedDicts but a sensible safety touch — `__forward_value__` is set lazily.

## Recommendation

Before merging:

1. Fix the Required/NotRequired symmetry check in `_unify_typeddict` (most important — the new tests assert the wrong behavior).
2. Either enforce invariance on mutable fields, or document explicitly that this codebase intentionally diverges from the spec to allow covariant field unification (and update the docstring).
3. Decide what to do for the `Mapping[K, V]` vs. TypedDict case — at minimum, add a regression test that exercises a TypedDict whose field types don't match `V`, to make the choice explicit.
4. Either fix `_get_typeddict_hints` to substitute through `__orig_bases__`, or change the inheritance-generic test to not depend on the leak.

Items 4 (`ReadOnly`), 6 (`nested_type` heuristic), 7 (`_Canonical` class name) are smaller and can be follow-ups.

kiranandcode:
Hmm, okay! will update!

eb8680 (approved):

PR #655, opened 2026-05-04
Add inversion from `weighted`

jfeser:
Depends on #653

### DIFF
--- a/effectful/ops/monoid.py
+++ b/effectful/ops/monoid.py
@@ -610,8 +610,8 @@ class ReduceDistributeCartesianProduct(ObjectInterpretation):
 
     More specifically, this transform implements the identity
     reduce(⨁, reduce(⨂, body2, {vv: v()}), {v: reduce(×, body1, S1)} ∪ S2)
-        = reduce(⨁, reduce(⨂, reduce(⨁, body2, {vv: v()}), S1), S2)
-            where × is the cartesian product and ⨂ distributes over ⨁.
+        = reduce(⨁, reduce(⨂, reduce(⨁, body2, {vv: body1}), S1), S2)
+    where × is the cartesian product and ⨂ distributes over ⨁.
 
     Note: This could be generalized to grouped inversion [2].
 
--- a/effectful/ops/monoid.py
+++ b/effectful/ops/monoid.py
@@ -10,18 +10,20 @@ from graphlib import TopologicalSorter
 from typing import Annotated, Any
 
 from effectful.internals.disjoint_set import DisjointSet
+from effectful.internals.runtime import interpreter
 from effectful.ops.semantics import coproduct, evaluate, fvsof, fwd, handler
 from effectful.ops.syntax import (
     ObjectInterpretation,
     Scoped,
     _NumberTerm,
     defdata,
+    deffn,
     implements,
     iter_,
     syntactic_eq,
     syntactic_hash,
 )
-from effectful.ops.types import Interpretation, NotHandled, Operation, Term
+from effectful.ops.types import Expr, Interpretation, NotHandled, Operation, Term
 
 # Note: The streams value type should be something like Iterable[T], but some of
 # our target stream types (e.g. jax.Array) are not subtypes of Iterable
@@ -80,9 +82,13 @@ class Monoid[T]:
     def _plus[S](self, *args: S) -> S:
         return typing.cast(S, functools.reduce(self.kernel, args, self.identity))
 
-    @_plus.register(Sequence)
+    @_plus.register(tuple)
     def _(self, *args):
-        return type(args[0])(self.plus(*vs) for vs in zip(*args, strict=True))
+        return tuple(self.plus(*vs) for vs in zip(*args, strict=True))
+
+    @_plus.register(Generator)
+    def _(self, *args):
+        return (self.plus(*vs) for vs in zip(*args, strict=True))
 
     @_plus.register(Mapping)
     def _(self, *args):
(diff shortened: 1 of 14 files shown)


effectful/ops/monoid.py
```diff
@@ -545,11 +561,153 @@ def reduce(self, monoid, body, streams):
...
+
+    topo = TopologicalSorter(succ)
+    topo.prepare()
+    return (
+        ({k: v for (k, v) in streams.items() if k != op}, op, streams[op])
+        for op in set(topo.get_ready()) | no_dependents
+    )
+
+
+def match_reduce(term: Term) -> tuple | None:
```
eb8680, line 595: This is a very strange function. Surely there's a more comprehensible way to write this, even if it ends up being more verbose.

effectful/ops/monoid.py
```diff
@@ -545,11 +561,153 @@ def reduce(self, monoid, body, streams):
         return fwd()
 
 
+def outer_stream(
```
eb8680: This doesn't seem to be used elsewhere?

effectful/ops/monoid.py
```diff
@@ -545,11 +561,153 @@ def reduce(self, monoid, body, streams):
...
+
+class ReduceDistributeCartesianProduct(ObjectInterpretation):
+    """Eliminates a reduce over a cartesian product.
+        ∑_x₁ ∑_x₂ ... ∑_xₙ ∏_i f(xᵢ) = ∏_i ∑_xᵢ f(xᵢ)
+    This transform is also called inversion in the lifting
+    literature (e.g. [1]).
+
+    More specifically, this transform implements the identity
+    reduce(⨁, reduce(⨂, body2, {vv: v()}), {v: reduce(×, body1, S1)} ∪ S2)
+        = reduce(⨁, reduce(⨂, reduce(⨁, body2, {vv: v()}), S1), S2)
```
eb8680: This equation is quite hard to parse and appears to be wrong (what binds `v` on the RHS?). Is it out of date wrt what is now happening in the rule?

tests/test_ops_monoid.py
```diff
@@ -121,7 +126,7 @@ def _apply_canonical(op, *args, **kwargs):
             *bindings.args, *bindings.kwargs.values()
         )
         if not all_bound:
-            return defdata(op, *args, **kwargs)
+            return _BaseTerm(op, *args, **kwargs)
```
eb8680, line 144: Why this change? It's not correct to construct and return a `_BaseTerm` explicitly, all terms should be constructed by `defdata`

tests/test_ops_syntax.py
```diff
@@ -489,7 +489,7 @@ def _(self, x: bool) -> bool:
     )
 
     assert isinstance(term_float, Term)
-    assert term_float.op.__name__ == "my_singledispatch"
+    assert term_float.op.__name__ == "MyClass_my_singledispatch"
```
eb8680: This is an implementation detail, it should just be removed from the test

tests/test_handlers_llm_provider.py
```diff
@@ -240,7 +240,7 @@ def test_agent_tool_names_are_valid_integration():
     agent = _ToolNameAgent()
     template = agent.ask
     tools = template.tools
-    expected_helper_tool_name = f"self__{agent.helper.__name__}"
+    expected_helper_tool_name = "self__helper"
```
eb8680, line 243: This name is an implementation detail, the test should check that `agent.helper` is in `tools.values()` or something similar so that it does not have to be changed like this in the future

effectful/ops/types.py
```diff
@@ -488,7 +488,10 @@ def _instance_op(instance, *args, **kwargs):
                     else:
                         return default_result
 
-                instance_op = self.define(types.MethodType(_instance_op, instance))
+                name = ("" if owner is None else f"{owner.__name__}_") + self.__name__
```
eb8680, line 491: Names of `Operation`s should be implementation details that don't affect behavior. Is this change fixing a bug you found? If so, maybe it should land on `master`

eb8680 (commented):
Definitely a nice simplification compared to the old version in #639 but can we add this to your hypothesis pipeline or otherwise test it much more extensively? Right now it's hard to say whether it's correct in general.

effectful/ops/monoid.py
jfeser: Agreed. Unfortunately, I'm not sure how to write this in a way that doesn't enumerate operations. The subclassing relation between operations is implicit, and we want to match all subclasses (and leave open the possibility that new ones could be created).

effectful/ops/types.py
jfeser: No, it's making the terms easier to read. Without this change, you get `reduce(reduce!1(reduce!2(...), ...), ...)` etc.

tests/test_ops_monoid.py
jfeser: It's critical to avoid the renaming behavior attached to `defdata`, because we're canonicalizing the term.

effectful/ops/monoid.py
jfeser: It should read:

```
reduce(⨁, reduce(⨂, body2, {vv: v()}), {v: reduce(×, body1, S1)} ∪ S2)
        = reduce(⨁, reduce(⨂, reduce(⨁, body2, {vv: body1}), S1), S2)
```

jfeser:
Note: the mapping of `reduce` over lists conflicts with the reduction rules for cartesian product. In particular, `CartesianProduct.reduce([], {i: []}) == CartesianProduct.identity` (because reduction over an empty domain is the identity) but the list rule produces `[]`. I've removed the mapping over lists, but we could consider doing something type directed to allow it (and similar rules) when the monoid is not list-typed.

effectful/ops/monoid.py
jfeser: Dropped. (It'll come back in #658).

tests/test_handlers_llm_provider.py
jfeser: I think this test is specifically for names, since there's some restrictions on them.

eb8680 (approved):

PR #658, opened 2026-05-06
Add `jax` array monoids and reduction rule

jfeser:
Blockd by #655


eb8680 (approved):
Looks reasonable. It would be nice to group some of the new one-off `ObjectInterpretation`s together into larger logical units but we can do that in a separate PR.

eb8680:
Should be good to merge once the tests are fixed.

PR #663, opened 2026-05-21
Add `delta` terms for array construction in `handlers.jax.monoid`

jfeser:


eb8680 (approved):
Merging as-is into `staging-weighted` to facilitate refactoring, per offline discussion

PR #665, opened 2026-05-26
Add weighted streams

jfeser:


effectful/ops/monoid.py
```diff
@@ -111,6 +109,21 @@ def reduce[A, B, U: Body](
             return self.plus(*new_reduces)
         raise NotHandled
 
+    @Operation.define
+    def weighted[Elem, A, B](
+        self,
+        stream: Annotated[Iterable, Scoped[B]],
+        var: Annotated[Operation[[], Elem], Scoped[A]],
+        weight: Annotated[T, Scoped[A | B]],
+    ) -> Annotated[Iterable, Scoped[B]]:
```
eb8680: Following up on the issues with type inference, the following test passes on `master`:
```py
def test_unify_iterable_callable():

    class Monoid[W]:
        type Stream[S] = collections.abc.Iterable[S]

        def weighted[T](
            self,
            stream: Stream[T],
            weight: collections.abc.Callable[[T], W]
        ) -> Stream[T]:
            raise NotImplementedError

    stream = [1., 2.]

    def weight[T](item: T) -> int:
        return 1

    sig_weighted = inspect.signature(Monoid().weighted)
    subs = unify(
        sig_weighted,
        sig_weighted.bind(
            nested_type(stream).value, nested_type(weight).value
        ),
    )
    return_type = substitute(sig_weighted.return_annotation, subs)
    assert return_type == Monoid[int].Stream[float]
    assert canonicalize(return_type) == collections.abc.Iterable[float]
```
It seems like the problem you ran into is with the concrete type provided for `stream` - you might just need to add a couple patterns for `nested_type` and/or `canonicalize` to make this work as expected.

jfeser: We can get pretty close to this typing rule, but there's a few sticking points:
1. We can't dispatch term construction on type aliases right now. We can either replace `Stream[T]` with `Iterable[T]` or modify the treatment of types during dispatch to look through aliases. I've chosen to do the first.
2. We have some rules that need to construct operations of type `T`, so we need to be able to infer `T`. I've added two new functions to enable this. `typeof_full` returns the full generic type of a term; `stream_element_type` takes a stream type and returns its element type. This deals with types like `jax.Array` that don't subclass `Iterable`.

jfeser: And another: #669

eb8680: > We can't dispatch term construction on type aliases right now

I'm not sure I understand the problem. The test above includes a `Stream` type alias that is eliminated by `canonicalize` during type inference. At no point is it ever necessary to construct a term such that `typeof(term) == Stream[<something>]`.

eb8680: > We have some rules that need to construct operations of type T, so we need to be able to infer T

Which rules? I'm not sure I understand the problem or use case here either. All types in our rule bodies should be fully concrete.

> This deals with types like jax.Array that don't subclass Iterable.

I thought we decided the simpler solution was to pass non-stream types through a smart constructor that returns an opaque stream type that we know will correctly `unify` with `Iterable`/`Stream`? If we are going to lie to the type checker we should do it from the outside, rather than adding unsound edge cases inside `unify` et al.

jfeser: > The test above includes a Stream type alias that is eliminated by canonicalize during type inference. At no point is it ever necessary to construct a term such that typeof(term) == Stream[<something>].

Your test works, but my tests don't. Maybe something is going wrong with `canonicalize`, but I definitely see attempts to create terms with type `Stream`.

> Which rules? I'm not sure I understand the problem or use case here either. All types in our rule bodies should be fully concrete.

`ReduceCartesianWeightedStream` constructs a reduction over the row, which needs an operation with stream-element type. The type should be concrete, but we have to find out what it is.

> I thought we decided the simpler solution was to pass non-stream types through a smart constructor that returns an opaque stream type that we know will correctly unify with Iterable/Stream?

I don't see how this solves the problem with `ReduceCartesianWeightedStream`, which still needs to know the element type.

eb8680: > Your test works, but my tests don't. Maybe something is going wrong with canonicalize, but I definitely see attempts to create terms with type Stream.

My test above doesn't construct any terms, it just shows that `unify` itself has no problem with this signature. Type aliases are supposed to be erased at runtime anyway, so I think it's fine to add a couple lines of code for a base case in `defdata` that eliminates them prior to term construction just like we already do with `Annotated`.

> ReduceCartesianWeightedStream constructs a reduction over the row, which needs an operation with stream-element type. The type should be concrete, but we have to find out what it is.

I think you can just apply a dummy operation with the right signature and let the existing type inference system handle the rest:
```py
@Operation.define
def next_(stream: Stream[T]) -> T:
  raise NotHandled

stream_elem_type = typeof(next_(s))
```

jfeser: I've added some code to resolve aliases and switched back from `Iterable[T]` to `Stream[T]`.

> I think you can just apply a dummy operation with the right signature and let the existing type inference system handle the rest

This can work if we extend unify to treat `jax.Array`s. `stream_element_type` avoids the need to change the type system in that way. I'm ambivalent about which we choose, but I lean towards not changing the type inference.

eb8680: > This can work if we extend unify to treat jax.Arrays

I don't think we should do this. Rather, we should not be using raw `jax.Array`s or other values as streams at all if they are not in fact instances of `Stream`. Things that are not `Stream`s need to pass through a smart constructor to become `Stream`s.

jfeser: `jax.Array` instances are instances of `Stream`, but `jax.Array` does not subclass `Stream`. I think we should avoid a wrapper op around arrays if we can, because we'll need to wrap and unwrap frequently.

eb8680: I really don't think we should be trying to monkey-patch typing behavior in upstream libraries, even if it is seemingly incorrect. If `jax.Array` is not a subclass of `Stream`, we shouldn't pretend otherwise. I am not seeing why we'd need to wrap and unwrap a given stream more than once, I'd expect wrapping to be limited to fully concrete arrays acting as streams which is a behavior I only see in one rule in `handlers.jax.monoid`.

eb8680 (approved):

PR #667, opened 2026-05-26
Fix `_DistributionTerm.expand` return-type annotation

### LINKED ISSUE #666: `_DistributionTerm.expand` and `.to_event` return `_ArrayTerm` instead of `Distribution`, breaking standard NumPyro vectorization idiom
Splitting out from #321's [comment](https://github.com/BasisResearch/effectful/issues/321#issuecomment-4529027264).
`_DistributionTerm.expand` is `@defop`-annotated to return `jax.Array` rather than `Distribution`. So a method chain like `Normal(mu_term, tau_term).expand([J]).to_event(1)` falls through to `_ArrayTerm` after `.expand([J])`, and the subsequent `.to_event(1)` then raises:

```
AttributeError: '_ArrayTerm' object has no attribute 'to_event'
```

`_DistributionTerm.to_event` has the same `@defop` return-annotation issue, so even if `.expand` were fixed, the next `.to_event` call would land in the same hole.

Here is a self-contained reproduction script:
```python
import jax, jax.numpy as jnp, jax.random as jr
import numpyro
from effectful.handlers.numpyro import Normal

mu = defop(jax.Array, name="mu")

def model():
    # mu() is a free-variable Term, so Normal(mu(), 1.0) is a _DistributionTerm.
    # Chaining .expand([3]).to_event(1) then hits the @defop return-type bug:
    # .expand returns _ArrayTerm, so .to_event lands on the wrong term type.
    numpyro.sample("theta", Normal(mu(), 1.0).expand([3]).to_event(1))

numpyro.infer.MCMC(numpyro.infer.NUTS(model),
                   num_warmup=10, num_samples=10,
                   ).run(jr.PRNGKey(0))
# AttributeError: '_ArrayTerm' object has no attribute 'to_event'
```

The intermediate `.expand([3])` result is an `_ArrayTerm` (because `_DistributionTerm.expand`'s `@defop` return type is `jax.Array`), so chaining `.to_event(1)` lands on the wrong term type. Trace excerpt:

```
File ".../effectful/handlers/numpyro.py", line ...
    def expand(self, ...) -> jax.Array:   # <-- should be Distribution
        ...
```

This blocks the textbook NumPyro vectorised hierarchical-model idiom for any model whose loc/scale arrive as effectful terms. The workaround is the per-element loop:

```python
for j in range(J):
    theta_j = numpyro.sample(f"theta_{j}", Normal(mu_term, tau_term))
```

which works but is O(J) Python overhead and prevents JAX-level vectorisation. The `.expand([J]).to_event(1)` form would be the natural one to write.

datvo06:
Closes #666.

Per [@eb8680's feedback](https://github.com/BasisResearch/effectful/pull/667) — sticking to NumPyro's `Distribution.expand` signature (`-> Distribution`) rather than monkey-patching it to `-> Self`. Reverted the `typing.Self` annotation attempt and the merge of #613. Back to registering a `defdata` case for `dist.Distribution` (per @jfeser's suggestion in the same thread).

**Three small changes in `effectful/handlers/numpyro.py`:**

1. `_DistributionTerm.expand` return annotation: `jax.Array` → `dist.Distribution`. Was routing `.expand([J])` through `_ArrayTerm`, breaking `.to_event(1)` chaining.

2. `_DistributionTerm._is_eager` recurses into `_DistributionTerm`-valued args (rather than treating them as non-eager via `is_eager_array`). Otherwise a fully-eager `D.to_event(k)` would be tagged non-eager and every downstream property/sample method raised.

3. `defdata.register(dist.Distribution)` mapping to `_DistributionMethodTerm`. Without it, `dist.Distribution` falls through to `_CallableTerm` (since `Distribution` defines `__call__`). The class overrides `_pos_base_dist` to materialise by `getattr(base, self._op.__name__)(*args[1:], **kwargs)`, so every inherited `@defop` (`batch_shape`, `event_shape`, `support`, `log_prob`, `sample`, …) resolves through the standard machinery against the real NumPyro distribution.

**Three regression tests** (each verified to fail on master):
- narrow chain construction (the exact `AttributeError` from #666)
- equational shape/support laws on `.expand` / `.to_event`
- end-to-end MCMC over `Normal(mu_term, 1.0).expand([3]).to_event(1)` with `mu_term` bound via `effectful.ops.semantics.handler`

`Beta(...).to_event(k)` / `Dirichlet(...).to_event(k)` parametrised cases stay xfail — the basic chain works, but composing `.expand_by` on top with indexed (named-dim) arrays hits a deeper indexed-dim integration matter, out of scope for this PR.

Lint clean (ruff / format / mypy).

### DIFF
--- a/effectful/handlers/numpyro.py
+++ b/effectful/handlers/numpyro.py
@@ -425,29 +425,16 @@ class _DistributionMethodTerm(_DistributionTerm):
 
     @functools.cached_property
     def _pos_base_dist(self) -> dist.Distribution:
-        # Materialise the deferred method op against a real NumPyro distribution
-        # built from the (now-eager) receiver. Inherited ``@defop`` methods
-        # (``batch_shape``, ``event_shape``, ``support``, ``log_prob``,
-        # ``sample``, ...) then resolve via the standard ``_DistributionTerm``
-        # machinery against this base.
+        # Materialise by recursively building the receiver's _pos_base_dist
+        # and invoking NumPyro's method of the same name on it. Works for any
+        # ``dist.Distribution``-returning method op without a per-op switch.
         receiver = self._args[0]
         base = (
             receiver._pos_base_dist
             if isinstance(receiver, _DistributionTerm)
             else receiver
         )
-        if self._op is _DistributionTerm.to_event:
-            n = (
-                self._args[1]
-                if len(self._args) > 1
-                else self._kwargs.get("reinterpreted_batch_ndims")
-            )
-            return base.to_event(n)
-        if self._op is _DistributionTerm.expand:
-            return base.expand(self._args[1])
-        raise NotImplementedError(
-            f"_DistributionMethodTerm._pos_base_dist for op {self._op}"
-        )
+        return getattr(base, self._op.__name__)(*self._args[1:], **self._kwargs)
 
 
 @defop
--- a/effectful/handlers/numpyro.py
+++ b/effectful/handlers/numpyro.py
@@ -396,6 +396,32 @@ to_event = _DistributionTerm.to_event
 expand = _DistributionTerm.expand
 
 
+@defdata.register(dist.Distribution)
+class _DistributionMethodTerm(_DistributionTerm):
+    """Fallback term for distribution-method ops whose return annotation is the
+    abstract ``dist.Distribution`` (e.g. ``expand`` and ``to_event`` when the
+    receiver is non-eager).
+
+    Without this registration, ``defdata`` dispatch on ``dist.Distribution`` falls
+    through to ``collections.abc.Callable`` (since ``Distribution`` defines
+    ``__call__``) and produces a ``_CallableTerm``, breaking chained method calls
+    like ``Normal(mu_term, 1.0).expand([J]).to_event(1)``. See #666.
+
+    The receiver of the deferred op is taken as the first positional argument; we
+    carry forward its distribution family so further chained methods remain
+    available on the resulting term.
+    """
+
+    def __init__(self, ty, op, *args, **kwargs):
+        receiver = args[0] if args else None
+        constr = (
+            receiver._constr
+            if isinstance(receiver, _DistributionTerm)
+            else dist.Distribution
+        )
+        super().__init__(constr, op, *args, **kwargs)
+
+
 @defop
 def Cauchy(loc=0.0, scale=1.0, **kwargs) -> dist.Cauchy:
     raise NotHandled
--- a/effectful/handlers/numpyro.py
+++ b/effectful/handlers/numpyro.py
@@ -226,10 +226,12 @@ class _DistributionTerm(dist.Distribution):
 
     @functools.cached_property
     def _is_eager(self) -> bool:
-        return all(
-            (not isinstance(x, Term) or is_eager_array(x))
-            for x in (*self.args, *self.kwargs.values())
-        )
+        def _arg_is_eager(x):
+            if isinstance(x, _DistributionTerm):
+                return x._is_eager
+            return not isinstance(x, Term) or is_eager_array(x)
+
+        return all(_arg_is_eager(x) for x in (*self.args, *self.kwargs.values()))
 
     @property
     def op(self):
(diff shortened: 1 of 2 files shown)


datvo06:
Not ready yet. Reproduction in #666 still fails, now as (`_CallableTerm` instead of `_ArrayTerm` previously). Also havent added the reproduction test.

datvo06:
Fixed `expand().to_event()`, but for full MCMC examples in #666, I'd need to update it, since NumPyro'd eventually need `mu()` actual value.

datvo06:
Ready now!

effectful/handlers/numpyro.py
```diff
@@ -396,6 +398,33 @@ def __str__(self):
 expand = _DistributionTerm.expand
 
 
+@defdata.register(dist.Distribution)
```
jfeser, line 401: I think the right fix for this is to merge #613, which will let us create a `_NormalTerm` for `Normal(mu()).expand(s)`.

tests/test_handlers_numpyro.py
```diff
@@ -929,3 +929,94 @@ def test_distribution_typeof():
...
+    mu = defop(jax.Array, name="mu")
+
+    expanded = dist.Normal(mu(), 1.0).expand([3])
+    assert isinstance(expanded, numpyro.distributions.Distribution)
+
+    chained = expanded.to_event(1)
+    assert isinstance(chained, numpyro.distributions.Distribution)
+
+
+def test_expand_to_event_shape_laws():
```
jfeser, line 954: This test doesn't expand a symbolic distribution, because `mu` is always concrete. Did it fail before?

datvo06:
Working on merging #613 and find a cleaner solution.

datvo06:
@jfeser I ran into this problem of using `unify` on an inherited type:
```python
  import typing                                                                                                                                                                                                                                                       
  from effectful.internals.unification import unify                                                                                                                                                                                                                   
  
  T = typing.TypeVar("T")                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                    
  class Foo: pass
  class Bar(Foo): pass

  assert unify(T, Bar) == {T: Bar}      # FAILS: gets {T: Foo}          
```
So whenever you unify an inherited type with T, it resolves to the parent class. Should we default to principal type or widening in unifying/type inferring? 
This is blocking the Self-based routing, because then `Normal(mu()).expand(s)` would result in `dist.Distribution` instead of `NormalTerm` via `Self -> dist.Normal` anddispatch falls back to         _CallableTerm.
Which raises another error.

jfeser:
That's definitely a bug.

eb8680:
I am not sure about the premise of this PR. The latest proposed typing rule is not consistent with the signature of [`numpyro.distributions.Distribution.expand`](https://github.com/pyro-ppl/numpyro/blob/master/numpyro/distributions/distribution.py#L499), which only promises to return a `Distribution`, not something of the same type as `self`:
```py
def expand(self, batch_shape: Sequence[int]) -> Distribution: ...
```
I'm missing context on what you're trying to accomplish but I think it's unwise in general to attempt to monkey-patch typing behavior of upstream libraries within `effectful` like this and I'd strongly recommend sticking to the NumPyro signature for `expand`.

jfeser:
Sorry, that's my mistake. In that case, we might register `_DistributionTerm` as a `defdata` case for `dist.Distribution`.

datvo06:
Oh got it. Will revert back to `_DistributionTerm` defdata

The context was that I was attempting to factor out the interfacing between effectful symbols and numpyro [here](https://github.com/datvo06/effectful_mcmc/blob/main/effectful_mcmc/primitives.py). So then you could do something like this:
```python
import jax.numpy as jnp
import jax.random as jr

from effectful.handlers.numpyro import Normal, HalfNormal
from effectful_mcmc import sample, MCMC, NUTS

def linear_regression(x, y):
    alpha = sample(Normal(0.0, 10.0))
    beta  = sample(Normal(0.0, 10.0))
    sigma = sample(HalfNormal(2.0))
    sample(Normal(alpha + beta * x, sigma), obs=y)
    return alpha, beta, sigma                      # return per-site effects

mcmc = MCMC(NUTS(linear_regression), num_warmup=500, num_samples=1000)
mcmc.run(jr.PRNGKey(0), x_data, y_data)

# Posteriors are keyed by the Operation behind each sample site;
# `term.op` recovers it. No string names to keep in sync.
alpha, beta, sigma = mcmc.model_return_value
samples = mcmc.get_samples()
print(float(samples[alpha.op].mean()),
      float(samples[beta.op].mean()),
      float(samples[sigma.op].mean()))
```
Which is a bit more like normal Python. I was thinking something like this could be factored out of RoboTL.

eb8680 (approved):

PR #668, opened 2026-05-27
Polymorphic Tool and Template signatures

### LINKED ISSUE #489: `Tool` and `Template` signatures should support polymorphic types

datvo06:
Closes #489.

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -36,7 +36,13 @@ from effectful.handlers.llm.template import (
     Tool,
     _is_recursive_signature,
 )
-from effectful.internals.unification import nested_type
+from effectful.internals.unification import (
+    Substitutions,
+    freetypevars,
+    nested_type,
+    substitute,
+    unify,
+)
 from effectful.ops.semantics import fwd, handler
 from effectful.ops.syntax import ObjectInterpretation, implements
 from effectful.ops.types import Operation
@@ -78,6 +84,31 @@ def _get_history() -> collections.OrderedDict[str, Message]:
     raise _NoActiveHistoryException(
         "No active message history. This operation should only be used within a handler that provides a message history."
     )
+
+
+@Operation.define
+def _get_active_type_subs() -> Substitutions:
+    """Active TypeVar substitution from the enclosing Template call, or {}
+    when no Template call is in flight.
+
+    Scoped by LiteLLMProvider._call via `handler({_get_active_type_subs: ...})`
+    for the lifetime of each call. Inner (nested or recursive) Template
+    calls read the outer binding, merge their own, and re-install.
+    """
+    return {}
+
+
+def _bound_for(tv: typing.Any) -> typing.Any | None:
+    """If `tv` is a TypeVar with a bound or constraints, return the
+    schema-usable type. Bound wins over constraints. Otherwise None.
+    """
+    if not isinstance(tv, typing.TypeVar):
+        return None
+    if tv.__bound__ is not None:
+        return tv.__bound__
+    if tv.__constraints__:
+        return typing.Union[tv.__constraints__]
+    return None
 
 
 def append_message(message: Message, last: bool = True) -> None:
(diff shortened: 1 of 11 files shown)


effectful/handlers/llm/completions.py
```diff
@@ -230,27 +263,61 @@ def call_assistant[T](
...
+                bound = _bound_for(tv)
+                if bound is not None:
+                    ann = substitute(ann, {tv: bound})
+            free |= freetypevars(ann)
+            specialized[name] = ann
+        if free:
+            raise TypeError(
+                f"Tool {k!r} has free TypeVar(s) {free} in its signature "
+                f"that are not pinned by the enclosing Template call. "
+                f"Resolve by either: (1) declaring the TypeVar at module "
```
eb8680, line 315: It seems there may be some misunderstanding of type variable scoping underpinning this PR. Type variables are bound by implicit for-all quantifiers at function or class definitions, not the lexical variable of a `TypeVar` instance; there is no such thing as a module-level type variable even in Python 3.12+.

For example, the following three pairs of functions have the same type variable scoping semantics, because function definitions bind their type variables:
```py
T = typing.TypeVar("T")
def f(x: T) -> T: ...  # T bound by a quantifier, not shared with g
def g(x: T) -> T: ...  # T bound by a quantifier, not shared with f

def f2[T](x: T) -> T: ...
def g2[T](x: T) -> T: ...

def f3[T2](x: T2) -> T2: ...
def g3[T3](x: T3) -> T3: ...
```
The only ways to share type variables are generic classes or closures:
```py
def f[T](x: T) -> T:
  def g(x: T) -> T: ...  # this T is bound by the quantifier on f
  def h[T](x: T) -> T: ...  # this T is fresh and bound by a new quantifier on h
  ...

class A[T]:
  def f(self, x: T) -> T: ...  # this T is bound by the quantifier on A
  def g(self, x: T) -> T: ...  #   and shared by f and g
```

eb8680 (commented):
Maybe I'm missing something, but I'm not seeing how this resolves or works around the problem described in https://github.com/BasisResearch/effectful/issues/489#issuecomment-3800638776

I think the only general solution to that is generate well-typed code for an expression that calls the tool, rather than attempting to decode the polymorphic arguments directly.

effectful/handlers/llm/completions.py
datvo06: Ah I see, I was thinking something like let-polymorphism. But it's clearer now. Let me update it.

eb8680: Update it how? If you fix the PR to be consistent with type variable scope semantics, you'll end up back at the status quo in `master`. I could be missing something but I'm currently not seeing a path forward for this approach.

datvo06: Yeah you're mostly right. I was planning to nuke the PR and redesign it based on the comment.
I'm thinkin of doing these two: 
1. Class-based sharing via `__orig_class__`. Tools and Templates defined as methods on a generic class (class Workflow[T]: ...); the class's T is the shared quantifier, recoverable at instantiation from __orig_class__. Lives in _collect_tools when walking methods of an Agent/Generic[T] instance.
2. For code generation, LLMs would emit Python expressions that call tools. To resolve Kiran's comment, we would let Python's type system resolving polymorphism at execute time.

I'll add more detailed plan below.

eb8680: Option 1 sounds complicated and potentially unsound. Option 2 can be made to work, but we're missing necessary machinery (e.g. a code execution tool) and it shouldn't be on by default as it involves executing LLM-generated code.

eb8680:
Closing per our discussion yesterday, any revised approach to #489 will be quite different from this PR and should probably be a fresh PR.

Issue #578, opened 2026-02-18
Add explicit handling of context in notebooks for type-checking of generated code


kiranandcode:
Issue #577 proposes dropping support for runtime types as this functionality is ad-hoc and incomplete, but this functionality is partially necessary for notebooks.

Instead of providing generic runtime type support which will be difficult to test and validate, we should add explicit special casing for notebooks that produces specific contexts when running in notebooks and reduces the slow down in those notebooks.

Issue #448, opened 2025-12-17
Support effect type annotations


jfeser:
Partially in response to #447, but effect types would be a general mechanism.

We should add an annotation that can appear on the return type of an effect. This annotation should specify which effects a handler is allowed to perform. After #424, templates will be effects, so will be able to carry this annotation.

```
@defop
def sample() -> float: ...

@defop
def print(msg: str) -> None: ...

@defop
def does_samples() -> Annotated[float, sample]: ...

@defop
def samples_and_prints() -> Annotated[float, sample, print]: ...

@defop
def recurses() -> Annotated[float, "recurses"]: ...
```

This is a (weak) form of effect typing, as in languages like Koka [1]. When a handler performs an effect that is not in this list, we should raise a `TypeError`. This error indicates a bug in the program that we could in principle catch statically with a more sophisticated type checker.

When an effect is unannotated, its handlers may perform any effect.

When handling `Template` calls, we should only provide the LLM with the tools that are listed. This gives template authors more control over which tools their templates can invoke.

Note: we need special case behavior for self-reference, because the effect symbol is not defined at the point that the annotation is created. We could use the name of the symbol or something like `typing.Self` as a marker.

Non-goals:
- Effect types for arguments.
- Generic effect types. We could allow annotations that relate the final effect type to an argument effect type, but this is more complex than we need to consider right now.

[1] https://koka-lang.github.io/koka/doc/book.html#sec-effect-types

eb8680:
I think with `Operation` subtyping (#349) now available we can use that as a mechanism for representing effect types:
```py
class Tool(Operation): ...

class MyToolType(Tool): ...

class MyOtherToolType(Tool): ...

@MyToolType.define
def my_tool(...): ...

@MyOtherToolType.define
def my_other_tool(...): ...

@Template.define
def prompt(x: str) -> Annotated[str, Uses[MyToolType]]: ...

@Template.define
def other_prompt(y: str) -> Annotated[str, Uses[MyOtherToolType]]: ...

@Template.define
def third_prompt(z: str) -> Annotated[str, Uses[MyToolType | MyOtherToolType]]: ...
```

eb8680:
> This is appealing, but I think users will want the option to specify individual effects without giving them a special class. I think it's reasonable to support both effects and effect classes in these annotations.

I think you could use `typing.Literal` to talk about individual effects idiomatically:
```py
@Operation.define
def my_tool(...): ...

@Operation.define
[...]

PR #670, opened 2026-05-30
Synthetic readers for lexical context (addresses #497)

### LINKED ISSUE #497: Injecting Symbols Definition into Prompt for Program Synthesis.
To match Python semantics, synthesized code should be aware of its surrounding context.
```python
class Finger:
     def some_specific_operations(self, ...):
          pass

class Hand:
     fingers: List[Finger]
       

@Template.define
 def generate_arm(t: Type) -> Arm:
      """Generate an Arm that works with Hand"
      raise NotHandled
```
Concretely, we want the LLM to be contextualized by the lexical context by its definition. In this case, if we only pass in the Hand as the type, then the resulting Arm might use `fingers` in the wrong way because it isn't aware of `some_specific_operations.`
Defining this `generate_arm` somewhere else should result in different semantics.

datvo06:
Closes #497.

Templates expose synthetic read-only tools for lexical symbols that are not already real tools. The LLM can call them to inspect lexical state on demand, instead of receiving the entire scope dumped into the system prompt. Picks up where #545 left off; #585 landed the system-prompt half partially.

**Example.** A Template defined with a module-global the LLM should be able to see:

```python
from effectful.handlers.llm import Template, LiteLLMProvider

_known_data = [10, 20, 30, 40, 50]

@Template.define
def report_sum() -> int:
    """Use the `_known_data` tool to read the list of numbers,
    then return their sum as an integer."""
    raise NotHandled

with handler(LiteLLMProvider()):
    result = report_sum()
    assert result == 150
```

The LLM sees a tool named `_known_data` whose description is "Read the value of lexical variable `_known_data` (type `list[int]`).". It calls the tool, receives the list, and returns the sum.

**Dispatch.** Used `functools.singledispatch` to define different readers:

Definition-readers fire for classes and functions. The reader takes `level: Literal["short", "full"]` and returns `str`. Short uses `pydoc.render_doc` (byte-equivalent to `help(obj)`); full uses `inspect.getsource`. The probe is `inspect.getsource` reachability; symbols whose source is unreachable (builtin C, REPL lambdas) are skipped.

Value-readers fire for everything else (the default branch). The reader is zero-arg and returns `env[name]` live. The probe is `pydantic.TypeAdapter(Encodable[T]).json_schema()`; any failure causes the symbol to be skipped silently. Catch is broad on purpose, because the probe chains through several third-party libraries (`nested_type`, `inspect.signature`, `typing.get_overloads`, Pydantic schema generation) any of which can crash on third-party objects.

`Tool`, `Agent`, and `ModuleType` values are registered to return `None`. Tool and Agent are already collected by the existing real-tool path; modules are too big to expose by default.

**System prompt.** A short static sentence is appended to `Template.__system_prompt__` so the LLM knows the read-only-readers category exists. The structured tools array carries per-tool semantics; the preface doesn't enumerate them.

**Tests.** Invariants pinned by the unit tests:

- A value-reader returns `env[name]` evaluated at call time, so mutations and rebinds are visible.
- Deleting `env[name]` after collection causes the reader to raise `KeyError` on invocation.
- The probe accepts a symbol iff it can be Pydantic-encoded (no false positives, no false negatives across the documented categories).
- Real Tools take precedence over same-named synthetic readers.
- Reader annotations carry no free TypeVars, so the polymorphic-Template substitution machinery from #668 is a no-op for them.
- Definition-reader short form contains the class/function docstring and method signatures; full form contains method bodies (proving it routes to `inspect.getsource` rather than `pydoc`).
- A class rebound in `env` after reader construction is reflected in subsequent definition-reader calls.
- Pydantic `BaseModel` subclasses route correctly through singledispatch despite having a non-`type` metaclass.
- Classes / functions whose source is unreachable are skipped at probe time, not at call time.
- TypeVars, modules, and `Box` values are filtered through their respective skip paths.
- `Template.__system_prompt__` contains the preface unconditionally.
- A plain value in the Template's lexical scope is callable as a synthetic reader through `Template.tools`.
- Two existing assertions in `test_template_method` and `test_template_method_nested_class` flip from `local_variable not in tools` to `in`, matching the new behavior.

Invariant pinned by the integration test: a Template defined alongside a module-global value, prompted to call the synthetic reader and report a derived value, produces the correct answer end-to-end through a real LLM. The fixture was recorded against gpt-4o-mini and replays cleanly.

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -5,9 +5,11 @@ import dataclasses
 import functools
 import inspect
 import json
+import pydoc
 import string
 import textwrap
 import traceback
+import types
 import typing
 import uuid
 
@@ -209,6 +211,112 @@ def _collect_tools(
     return result
 
 
+def _build_definition_reader(
+    value: typing.Any, name: str, env: collections.abc.Mapping[str, typing.Any]
+) -> Tool | None:
+    """Build a Tool that returns the source / help() output of a class or
+    function.  Probe is `inspect.getsource` reachability; symbols whose
+    source is unreachable (builtin C, REPL lambdas, etc.) are skipped."""
+    try:
+        inspect.getsource(value)
+    except (OSError, TypeError):
+        return None
+
+    kind = "class" if inspect.isclass(value) else "function"
+
+    def body(level: typing.Literal["short", "full"] = "short") -> str:
+        obj = env[name]
+        if level == "short":
+            return pydoc.render_doc(obj, renderer=pydoc.plaintext)
+        return inspect.getsource(obj)
+
+    body.__name__ = name
+    body.__doc__ = (
+        f"Read the definition of `{name}` (a {kind}). "
+        f'`level="short"` (default) returns `help({name})` output; '
+        f'`level="full"` returns `inspect.getsource({name})`. '
+        f"Calls must include `level` explicitly under OpenAI strict mode."
+    )
+    body.__annotations__ = {
+        "level": typing.Literal["short", "full"],
+        "return": str,
+    }
+    return Tool.define(body)
+
+
+@functools.singledispatch
+def _build_synthetic_reader(
+    value: typing.Any,
+    name: str,
+    env: collections.abc.Mapping[str, typing.Any],
+) -> Tool | None:
+    """Build a synthetic read-only Tool for a lexical symbol.
+
+    Default branch: value-reader. Probe is the §2c Encodable schema check;
+    on failure (unencodable value), return None and let the symbol be
+    skipped silently.
+
+    Registrations route classes and functions through the
+    definition-reader path (`_build_definition_reader`), and route Tool,
+    Agent, and module values to a no-op (already collected by
+    `_collect_tools` or intentionally excluded).
+    """
+    try:
+        inferred = nested_type(value).value
+        adapter = pydantic.TypeAdapter(Encodable[inferred])
+        adapter.json_schema()
+    except Exception:
+        # The probe chains through several third-party libraries
+        # (nested_type → inspect.signature → typing.get_overloads →
+        # Pydantic schema generation). Any failure means "this symbol
+        # cannot be exposed as a synthetic reader" — the contract is
+        # skip-on-probe-failure, so we catch broadly. The cost of a
+        # too-narrow catch is a crash mid-call_assistant; the cost of a
+        # too-broad catch is silently skipping a symbol that might have
+        # worked.
+        return None
+
+    def body():
+        return env[name]
+
+    body.__name__ = name
+    body.__doc__ = f"Read the value of lexical variable `{name}` (type `{inferred}`)."
+    body.__annotations__ = {"return": inferred}
+    return Tool.define(body)
+
+
+_build_synthetic_reader.register(type, _build_definition_reader)
+_build_synthetic_reader.register(types.FunctionType, _build_definition_reader)
+_build_synthetic_reader.register(types.BuiltinFunctionType, _build_definition_reader)
+_build_synthetic_reader.register(types.MethodType, _build_definition_reader)
+
+
+@_build_synthetic_reader.register(types.ModuleType)
+@_build_synthetic_reader.register(Tool)
+@_build_synthetic_reader.register(Agent)
+def _no_synthetic_reader(value, name, env):
+    return None
+
+
+def _collect_synthetic_readers(
+    env: collections.abc.Mapping[str, typing.Any],
+    already_collected: collections.abc.Set[str],
+) -> collections.abc.Mapping[str, Tool]:
+    """Synthetic readers for lexical symbols not already covered by
+    `_collect_tools`. Skips dunder names and any name already in
+    `already_collected`."""
+    result: dict[str, Tool] = {}
+    for name, obj in env.items():
+        if name in already_collected:
+            continue
+        if name.startswith("__"):
+            continue
+        tool = _build_synthetic_reader(obj, name, env)
+        if tool is not None:
+            result[name] = tool
+    return result
+
+
 @Operation.define
 @functools.wraps(litellm.completion)
 def completion(*args, **kwargs) -> typing.Any:
@@ -243,7 +351,12 @@ def call_assistant[T](
         ResultDecodingError: If the result cannot be decoded. The error
             includes the raw assistant message for retry handling.
     """
-    tools = _collect_tools(env)
+    tools = dict(_collect_tools(env))
+    # Add synthetic readers for non-Tool lexical symbols.  These mirror the
+    # bound-args layer that LiteLLMProvider._call adds to env, so the LLM
+    # sees readers for the Template's arguments alongside other lexical
+    # context.
+    tools.update(_collect_synthetic_readers(env, set(tools)))
     tool_specs = {
         k: typing.cast(
             pydantic.TypeAdapter[typing.Any],
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -8,7 +8,6 @@ import json
 import string
 import textwrap
 import traceback
-import types
 import typing
 import uuid
 
@@ -179,88 +178,33 @@ class ToolCallExecutionError[E: Exception, T](DecodingError[E]):
 type MessageResult[T] = tuple[Message, typing.Sequence[DecodedToolCall], T | None]
 
 
-class _LexicalVariableTool[T](Tool[[], T]):
-    """A Tool that returns the current value of a variable captured from
-    a Template's lexical context.
+def _define_lexical_reader(
+    env: collections.abc.Mapping[str, typing.Any], *, name: str
+) -> "Tool[[], typing.Any]":
+    """Construct a synthetic reader Tool for `env[name]`.
 
-    Callers must gate construction on `_is_synthetic_reader_eligible`;
-    this classmethod assumes the value has already been deemed
-    schema-encodable and does no further checks.
+    Raises if the value's `Encodable[nested_type(value)]` schema cannot
+    be generated.  The caller is responsible for catching the failure
+    and deciding whether to skip the symbol.
     """
+    value = env[name]
+    typ: typing.Any = nested_type(value).value
+    # Probe schema generation; raises if `Encodable[typ]` is not implemented.
+    pydantic.TypeAdapter(Encodable[typ]).json_schema()
 
-    @classmethod
-    def define(
-        cls,
-        env: collections.abc.Mapping[str, typing.Any],
-        *,
-        name: str,
-        **kwargs,
-    ) -> "Tool[[], typing.Any]":
-        value = env[name]
-        typ: typing.Any = nested_type(value).value
+    def tool_fn():
+        return env[name]
 
-        def tool_fn():
-            return env[name]
-
-        tool_fn.__name__ = name
-        tool_fn.__qualname__ = name
-        tool_fn.__module__ = type(value).__module__
-        tool_fn.__doc__ = (
-            f"Reads the value of lexical variable `{name}` from the "
-            f"enclosing scope where this Template was defined.  Takes "
-            f"no arguments; returns the current value."
-        )
-        tool_fn.__annotations__ = {"return": typ}
-        return super().define(tool_fn, **kwargs)
-
-
-# Exceptions raised by the eligibility probe in
-# `_is_synthetic_reader_eligible`.  `TypeError` / `AttributeError` /
-# `NameError` cover empirically-observed `nested_type` failures on
-# pathological values (`pytest.mark.parametrize`, method descriptors,
-# Operations with unresolved forward-refs); see #673.
-_SYNTHETIC_READER_PROBE_FAILURES: tuple[type[BaseException], ...] = (
-    pydantic.errors.PydanticSchemaGenerationError,
-    pydantic.errors.PydanticInvalidForJsonSchema,
-    pydantic.errors.PydanticUserError,
-    TypeError,
-    AttributeError,
-    NameError,
-)
-
-
-# Concrete types that are never wrapped as synthetic readers regardless
-# of whether `Encodable[T]` would happen to schematise them.  Class
-# objects with annotated `__init__` route through `nested_type`'s
-# Callable branch and slip past `Encodable[type]`; modules / functions
-# / methods / builtins / Agent instances would otherwise be wrapped as
-# `Callable[Args, Return]` synthesis schemas.
-_NON_READER_TYPES: tuple[type, ...] = (
-    type,
-    types.ModuleType,
-    types.FunctionType,
-    types.MethodType,
-    types.BuiltinFunctionType,
-    Agent,
-    Tool,
-)
-
-
-def _is_synthetic_reader_eligible(value: typing.Any) -> bool:
-    """Decide whether `value` should be exposed as a synthetic reader.
-
-    A value is eligible iff it is not an instance of any
-    `_NON_READER_TYPES` class AND its `Encodable[nested_type(value)]`
-    schema can be generated.
-    """
-    if isinstance(value, _NON_READER_TYPES):
-        return False
-    try:
-        typ = nested_type(value).value
-        pydantic.TypeAdapter(Encodable[typ]).json_schema()
-    except _SYNTHETIC_READER_PROBE_FAILURES:
-        return False
-    return True
+    tool_fn.__name__ = name
+    tool_fn.__qualname__ = name
+    tool_fn.__module__ = type(value).__module__
+    tool_fn.__doc__ = (
+        f"Reads the value of lexical variable `{name}` from the "
+        f"enclosing scope where this Template was defined.  Takes "
+        f"no arguments; returns the current value."
+    )
+    tool_fn.__annotations__ = {"return": typ}
+    return Tool.define(tool_fn)
 
 
 def _collect_tools(
@@ -278,12 +222,21 @@ def _collect_tools(
                 for attr_name in vars(cls):
                     if isinstance(getattr(obj, attr_name), Tool):
                         result[f"{name}__{attr_name}"] = getattr(obj, attr_name)
-        elif (
-            name.isidentifier()
-            and not name.startswith("__")
-            and _is_synthetic_reader_eligible(obj)
-        ):
-            result[name] = _LexicalVariableTool.define(env, name=name)
+        elif name.isidentifier() and not name.startswith("__"):
+            try:
+                result[name] = _define_lexical_reader(env, name=name)
+            # `TypeError`/`AttributeError`/`NameError` absorb `nested_type`
+            # bugs on pathological values (MarkDecorator, method descriptors,
+            # Operations with unresolved forward-refs); see #673.
+            except (
+                pydantic.errors.PydanticSchemaGenerationError,
+                pydantic.errors.PydanticInvalidForJsonSchema,
+                pydantic.errors.PydanticUserError,
+                TypeError,
+                AttributeError,
+                NameError,
+            ):
+                continue
 
     # Same Tool can appear under multiple names when visible both in the
     # enclosing scope and via an Agent instance's MRO.  Keep only the
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -220,39 +220,20 @@ class _LexicalVariableTool[T](Tool[[], T]):
 
 
 @Operation.define
-def expose_lexical_readers() -> bool:
-    """Effect controlling whether `_collect_tools` builds synthetic
-    read-only Tools for non-Tool/Template values in a Template's
-    lexical scope.
-
-    Default behaviour is *off*: only real Tools/Templates/Agents reach
-    the LLM, and the lexical context is invisible.  Install
-    `LexicalReaders` to flip it on for the call-site where the LLM
-    should be able to inspect closure state.
-    """
-    return False
-
-
-class LexicalReaders(ObjectInterpretation):
-    """Handler that enables synthetic lexical-reader generation in
-    `_collect_tools`.  Each plain value in a Template's lexical context
-    becomes a zero-argument Tool that returns the captured value.
-    """
-
-    @implements(expose_lexical_readers)
-    def _enabled(self) -> bool:
-        return True
-
-
-def _collect_tools(
+def collect_tools(
     env: collections.abc.Mapping[str, typing.Any],
 ) -> collections.abc.Mapping[str, Tool]:
-    """Operations and Templates available as tools.  When
-    `expose_lexical_readers` is on (see :class:`LexicalReaders`),
-    plain values in the lexical context are also wrapped as synthetic
-    read-only tools."""
+    """Return the tools available to a Template given its lexical context.
+
+    Default rule: real `Tool` and `Template` values bound directly in
+    `env`, plus `Tool` methods discovered through the MRO of any
+    `Agent` instance in `env`.  Same-Tool-under-different-names is
+    deduped so each Tool appears exactly once.
+
+    Handlers (see :class:`LexicalReaders`) may override this to add
+    synthetic readers, hide tools, etc.
+    """
     result: dict[str, Tool] = {}
-    readers_on = expose_lexical_readers()
 
     for name, obj in env.items():
         if isinstance(obj, Tool | Template):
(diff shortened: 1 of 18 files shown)


datvo06:
**Filtering*, currently, we don't do any filtering, so for example, agent *could* fetch the API keys through env var if it's in the same lexical context.

```python
API_KEY = ...
@Template.define
def generate_some_thing(...) -> str:
   raise NotHandled
```

Other than that, we are also including:
- ABCs with pure-Python source (`collections.abc.Mapping`, `Iterable`, `Callable`), Stdlib helpers with reachable source (`os.path.join`, `pathlib.Path`) become definition-readers.
- Pydantic-serializable user values, including `re.Pattern` and `pathlib.PosixPath` instances, become value-readers.
- Stdlib builtin-C types (`int`, `list`, `len`) fail `inspect.getsource` and get skipped.
- TypeVars, `Term`, `Operation`, `Box` values fail the Encodable probe and get skipped.

Thinking of adding `hide=[...]` or `expose=[...]` as a monkey patch if needed but effect typing #448 will be a cleaner solution.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+        # The cost of a too-narrow catch is a crash mid-call_assistant;
+        # the cost of a too-broad catch is silently skipping a symbol
+        # that might have worked.
+        return None
+
+    def body():
+        return env[name]
+
+    body.__name__ = name
+    body.__doc__ = f"Read the value of lexical variable `{name}` (type `{inferred}`)."
```
eb8680: The type should already be part of the tool schema via the return type annotation, you shouldn't need to repeat it in the docstring.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
     return result
 
 
+def _build_definition_reader(
```
eb8680: This seems like a duplicate of `Encodable` for `Callable`.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+    function.  Probe is `inspect.getsource` reachability; symbols whose
+    source is unreachable (builtin C, REPL lambdas, etc.) are skipped."""
+    try:
+        inspect.getsource(value)
+    except (OSError, TypeError):
+        return None
+
+    kind = "class" if inspect.isclass(value) else "function"
+
+    def body(level: typing.Literal["short", "full"] = "short") -> str:
```
eb8680: This `level` toggle is an interesting idea, but it should probably enter as part of `Encodable`, if at all. I would suggest removing this distinction and the new `"short"` path from the PR for now since it seems like an optimization and instead deferring to whatever the current behavior is for `Encodable`.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+    )
+    body.__annotations__ = {
+        "level": typing.Literal["short", "full"],
+        "return": str,
+    }
+    return Tool.define(body)
+
+
+@functools.singledispatch
+def _build_synthetic_reader(
```
eb8680: Making this a `singledispatch` function seems like it's going to duplicate a lot of logic that should really just be in `Encodable`. I'd maybe also make this a special internal-only `Tool` subclass in case it's useful elsewhere in the code to distinguish these special tools from regular ones:
```py
class _LexicalVariableTool[T](Tool[[], T]):
    """A Tool wrapper for a variable captured from the lexical context. This allows
    variables to be automatically available as tools without explicit wrapping.

    The tool takes no arguments and returns the value of the variable when called.
    """
    @classmethod
    def define(cls, value: T, *, name: str, **kwargs) -> Tool[[], T]:
        assert name.isidentifier()
        assert not isinstance(value, Tool)
        typ: type[T] = nested_type(value).value  # or maybe just type(value)?
        assert pydantic.TypeAdapter(Encodable[typ]).json_schema()
        tool_fn = lambda: value
        tool_fn.__name__ = name
        tool_fn.__qualname__ = name
        tool_fn.__module__ = value.__module__
        tool_fn.__doc__ = f"""Reads value of lexical variable `{name}`"""
        tool_fn.__annotations__.update({"return": typ})
        return super().define(tool_fn, name=name, **kwargs)
```

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+    def body():
+        return env[name]
+
+    body.__name__ = name
+    body.__doc__ = f"Read the value of lexical variable `{name}` (type `{inferred}`)."
+    body.__annotations__ = {"return": inferred}
+    return Tool.define(body)
+
+
+_build_synthetic_reader.register(type, _build_definition_reader)
```
eb8680: Seems like this should be subsumed by `Encodable[type]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+_build_synthetic_reader.register(types.FunctionType, _build_definition_reader)
```
eb8680: Seems like this should be subsumed by `Encodable[types.FunctionType]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+_build_synthetic_reader.register(types.BuiltinFunctionType, _build_definition_reader)
```
eb8680: Seems like this should be subsumed by `Encodable[types.BuiltinFunctionType]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+_build_synthetic_reader.register(types.MethodType, _build_definition_reader)
```
eb8680: Seems like this should be subsumed by `Encodable[types.MethodType]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+
+
+@_build_synthetic_reader.register(types.ModuleType)
```
eb8680: Seems like this should be subsumed by `Encodable[types.ModuleType]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+@_build_synthetic_reader.register(Tool)
```
eb8680: This seems like something that should be hard-coded into the constructor - we know `Tool`s should never be wrapped in this way.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+@_build_synthetic_reader.register(Agent)
```
eb8680: Seems like this should be subsumed by `Encodable[Agent]`

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+    on failure (unencodable value), return None and let the symbol be
+    skipped silently.
+
+    Registrations route classes and functions through the
+    definition-reader path (`_build_definition_reader`), and route Tool,
+    Agent, and module values to a no-op (already collected by
+    `_collect_tools` or intentionally excluded).
+    """
+    try:
+        inferred: typing.Any = nested_type(value).value
```
eb8680: `nested_type` is supposed to return `Box(type(value))` rather than raise a `TypeError` whenever it gets a value it doesn't understand - if it raises an error that's most likely a bug in `nested_type` and it should fail loudly.

effectful/handlers/llm/completions.py
```diff
@@ -209,6 +211,113 @@ def _collect_tools(
...
+        adapter: pydantic.TypeAdapter[typing.Any] = pydantic.TypeAdapter(
+            Encodable[inferred]
+        )
+        adapter.json_schema()
+    except Exception:
+        # The probe chains through several third-party libraries
+        # (nested_type, inspect.signature, typing.get_overloads,
+        # Pydantic schema generation). Any failure means "this symbol
```
eb8680: I think schema generation is the only thing we'd want to fail silently, and I think it should raise a more specific Pydantic error type that we can catch here rather than just catching any `Exception`.

tests/test_handlers_llm_provider.py
```diff
@@ -2158,3 +2158,36 @@ def _completion(self, model, messages=None, **kwargs):
...
+
+
+# ---------------------------------------------------------------------------
+# Synthetic readers — integration (PR #545 finish-up)
+# ---------------------------------------------------------------------------
+
+
+# Module-level binding the LLM will be asked to inspect via a synthetic
+# reader.  The reader's name in the tool list is `_known_data`.
+_known_data = [10, 20, 30, 40, 50]
```
eb8680: nit: put this in the test function just above `report_sum`, there's no reason for it to be a global.

effectful/handlers/llm/template.py
```diff
@@ -11,6 +11,12 @@
 
 from effectful.ops.types import Annotation, Operation
 
+_LEXICAL_READERS_PREFACE = (
```
eb8680: This information seems like it should be part of the docstring of each generated lexical variable tool, rather than tacked onto the system prompt once. That way there's no ambiguity from the model's perspective about what each individual tool does.

effectful/handlers/llm/completions.py
```diff
@@ -243,7 +352,12 @@ def call_assistant[T](
         ResultDecodingError: If the result cannot be decoded. The error
             includes the raw assistant message for retry handling.
     """
-    tools = _collect_tools(env)
+    tools = dict(_collect_tools(env))
+    # Add synthetic readers for non-Tool lexical symbols.  These mirror the
+    # bound-args layer that LiteLLMProvider._call adds to env, so the LLM
+    # sees readers for the Template's arguments alongside other lexical
+    # context.
+    tools.update(_collect_synthetic_readers(env, set(tools)))
```
eb8680: Rather than making `_collect_synthetic_readers` a separate step, I'd inline its body inside `_collect_tools` so that we benefit from the deduplication and sorting that already happens there.

eb8680 (commented):
Thanks for taking a pass at this. I think it can be quite a bit simpler if you defer most behavior to `Encodable`. That even includes types like `type` and `types.ModuleType` that are currently missing `Encodable` implementations - they should still trigger Pydantic schema generation errors, which you can catch and use to skip tool generation.

effectful/handlers/llm/completions.py
eb8680: Another benefit of doing this is that the class docstring is a natural home for information about how to use such tools (e.g. during synthesis #497 ) that you'd like to inject into the system prompt.

eb8680:
Also, I don't think this closes #497 in its current form. I don't see any tests for context-sensitivity during synthesis and there's currently nothing to indicate to the LLM that these are lexical variables that are available in generated code. We could leave that behavior for a followup PR to keep this one tractable, although I don't think including it would require much more library code, just more testing.

eb8680:
This is probably also going to cause problems with notebooks because of all the stuff they inject into `globals` (cf #578) - maybe we should gate the behavior behind a handler, or just disable it for now when we detect that a `Template.__context__` came from a notebook?

effectful/handlers/llm/completions.py
datvo06: yeah, I briefly considered that but thought that we could want different texts for symbols, I think what you said make sense, will fold into the help tool and make it go through encodables instead.

effectful/handlers/llm/completions.py
datvo06: Got it. Will remove and defer to next PR if we run into trouble.

datvo06:
Update: dropped the per-type singledispatch and the definition-reader path. Now one free function `_define_lexical_reader` builds the tool and `_collect_tools` catches probe failures inline. In detail:

- Annotated callables flow through as Callable-synthesis tools via `_pydantic_callable`. Classes, methods, builtins do the same. No per-type code path.
- Filter is `name.isidentifier()` only - module dunders flow through if encodable.
- Catching Pydantic schema errors so that unencodable things get passed. One caveat is that I had to include these extra errors `TypeError`/`AttributeError`/`NameError` (See #673). 

One synthesis-oriented test is masked skipping now pending #674. That's likely what happened in RoboTL's https://github.com/BasisResearch/RoboTL/pull/395 and https://github.com/BasisResearch/RoboTL/pull/444 that made mypy type check crash. Will unskip that after #675 and will also fix downstream.

effectful/handlers/llm/completions.py
```diff
@@ -178,29 +178,69 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
 type MessageResult[T] = tuple[Message, typing.Sequence[DecodedToolCall], T | None]
 
 
+def _define_lexical_reader(
```
eb8680: nit: make this a subtype of `Tool` with this as a `define` method

effectful/handlers/llm/completions.py
```diff
@@ -178,29 +178,69 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
...
+    env: collections.abc.Mapping[str, typing.Any], *, name: str
```
eb8680: Just take the value here, you don't need the whole environment. These tools are constructed anew in every `call_assistant` call, so there's no advantage to carrying around the environment in the closure.

effectful/handlers/llm/completions.py
```diff
@@ -178,29 +178,69 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
...
+            # `TypeError`/`AttributeError`/`NameError` absorb `nested_type`
+            # bugs on pathological values (MarkDecorator, method descriptors,
+            # Operations with unresolved forward-refs); see #673.
+            except (
+                pydantic.errors.PydanticSchemaGenerationError,
+                pydantic.errors.PydanticInvalidForJsonSchema,
+                pydantic.errors.PydanticUserError,
+                TypeError,
+                AttributeError,
+                NameError,
```
eb8680: If these last three cases are from #673 it would probably be better to fix that first before landing this PR so we don't forget to remove them afterward.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    abort tool collection; the probe catches the `AttributeError` from
+    `nested_type`'s Callable branch (`typing.get_overloads` accesses
+    `__qualname__` which `MarkDecorator` lacks)."""
+    env = {"mark": pytest.mark.parametrize, "regular": 42}
+    result = _collect_tools(env)
+    assert "mark" not in result
+    assert "regular" in result
+
+
+def test_system_prompt_has_no_lexical_readers_preface():
```
eb8680: This test is weak and brittle and should be removed given that it's very likely to break in the future

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+        "d": {"k": 1},
+        "model": _SimpleModel(x=1, y="hi"),
+    }
+    result = _collect_tools(env)
+    assert {"x", "s", "lst", "d", "model"} <= set(result)
+    assert result["x"]() == 1
+    assert result["model"]() == env["model"]
+
+
+def test_lexical_reader_skips_marker_objects():
```
eb8680: Isn't this going to fail after #673 is addressed? If so, it should be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+        def t(self) -> int:
+            """Doc."""
+            return 7
+
+    inst = _A()
+    env = {"a": inst}
+    result = _collect_tools(env)
+    assert "a" not in result
+    # The MRO walk picks up the contained Tool.
+    assert any(k.startswith("a__") for k in result)
```
eb8680: This is checking an implementation detail that is likely to change. Prefer checking values over names whenever possible (`inst.t in result.values()`).

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    explicitly avoid binding these names or filter them at the handler
+    layer (e.g., `Template.tools` is not the place to second-guess what
+    Encodable accepts).
+    """
+    value = make_value()
+    env = {name: value}
+    assert name in _collect_tools(env)
+
+
+def test_collect_tools_skips_modules():
```
eb8680: This is checking an implementation detail, not a semantic invariant. It's a weak test that should be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+
+
+def test_collect_tools_skips_unencodable_values():
+    """`_collect_tools` catches probe failures and omits the symbol."""
+    for name, value in _PROBE_FAIL_CASES:
+        env = {name: value}
+        assert name not in _collect_tools(env)
+
+
+def test_collect_tools_real_tools_take_precedence_over_value_readers():
```
eb8680: This is checking an invariant (`Tool`s don't get re-encoded as lexical variables) that should actually be enforced by a runtime assertion inside `_collect_tools`. I'm not sure the test is adding any value.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+)
+def test_define_lexical_reader_raises_for_unencodable(name, value):
+    """Schema-generation failures propagate from `_define_lexical_reader`;
+    the call site decides whether to skip."""
+    env = {name: value}
+    with pytest.raises(Exception):
+        _define_lexical_reader(env, name=name)
+
+
+def test_collect_tools_skips_unencodable_values():
```
eb8680: This test is weak, not specific to `_collect_tools` and should be removed; it's just checking behavior of `Encodable` that is very likely to change in the future.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    tool = _define_lexical_reader(env, name=name)
+    assert tool() == value
+
+
+@pytest.mark.parametrize(
+    "name,value",
+    _PROBE_FAIL_CASES,
+    ids=lambda x: x[0] if isinstance(x, tuple) else None,
+)
+def test_define_lexical_reader_raises_for_unencodable(name, value):
```
eb8680: This test is just checking behavior of `Encodable` that is very likely to change in the future and should be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+
+@pytest.mark.parametrize(
+    "name,value", _PROBE_OK_CASES, ids=lambda x: x[0] if isinstance(x, tuple) else None
+)
+def test_define_lexical_reader_returns_value(name, value):
+    """`_define_lexical_reader` builds a Tool when `Encodable[T]`
+    schema generates; calling it returns the live value."""
+    env = {name: value}
+    tool = _define_lexical_reader(env, name=name)
+    assert tool() == value
```
eb8680: Should be strengthened to `tool() is value`

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    body reads `env[name]` at call time, not a snapshot."""
+    env: dict = {"x": 42}
+    tool = _define_lexical_reader(env, name="x")
+    assert tool() == 42
+    env["x"] = 99
+    assert tool() == 99
+
+
+def test_synthetic_reader_raises_when_name_deleted():
+    """If env[name] is deleted between collection and call, the reader
```
eb8680: I don't think this scenario should be possible? `Template.__context__` is a proxy for the context - if something is added or removed between the time when a `Template` is defined and when it is invoked, that should be automatically reflected by `Template.__context__`. I'd remove this test entirely, we should never be calling `_define_lexical_reader` on arbitrary environments that are not derived from `Template.__context__` so this is undefined behavior.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    """A reader returns whatever env[name] currently is. Mutation
+    after reader construction propagates."""
+    env: dict = {"x": [1, 2, 3]}
+    tool = _define_lexical_reader(env, name="x")
+    assert tool() == [1, 2, 3]
+    env["x"].append(4)
+    assert tool() == [1, 2, 3, 4]
+
+
+def test_synthetic_reader_rebind():
```
eb8680: This is checking a property that should probably not hold. At least in the current implementation, these tools are ephemeral artifacts derived from a `Template.__context__` that is the real source of truth and do not persist beyond a single `call_assistant` call, so they'd never be invoked twice and this scenario should be impossible. The test should just be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+        """Doc."""
+        return 1
+
+    env = collections.ChainMap({"shared": shared, "shared_value": 42})
+    result = _collect_tools(env)
+    assert result["shared"] is shared
+    assert result["shared_value"]() == 42
+
+
+def test_synthetic_reader_annotation_has_no_free_typevars():
```
eb8680: This is checking a property of `nested_type` that is not specific to this use case and may not hold in general (e.g. polymorphic function types may be allowed). It should be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+        "x": 1,
+        "s": "hello",
+        "lst": [1, 2, 3],
+        "d": {"k": 1},
+        "model": _SimpleModel(x=1, y="hi"),
+    }
+    result = _collect_tools(env)
+    assert {"x", "s", "lst", "d", "model"} <= set(result)
+    assert result["x"]() == 1
+    assert result["model"]() == env["model"]
```
eb8680: Should check identity, not equality

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1534,299 @@ def test_tool_forward_ref():
...
+    def t(x: int) -> int:
+        """Doc."""
+        raise NotHandled
+
+    # Module docstring may be present, but the old preface phrasing
+    # ("You also have access to read-only tools") is not.
+    assert "read-only tools for inspecting the lexical" not in t.__system_prompt__
+
+
+def test_lexical_reader_doc_mentions_name():
```
eb8680: This is checking an implementation detail that may change - as with the type, we may not actually want/need to include the name in the docstring, since it's already in the tool schema via the tool name. It is brittle, weak and should be removed.

tests/test_handlers_llm_tool_calling_poem.py
```diff
@@ -92,7 +92,10 @@ def generate_good_poem(topic: str) -> Poem:
     Keep iterating until evaluate_poem_tool returns GOOD.
     Return your final poem as JSON with 'content' and 'form' fields.
 
-    Do not call the 'generate_good_poem' tool.
+    Do not call any tool other than evaluate_poem_tool. In particular,
+    do not call generate_good_poem and do not call any read-only lexical
```
eb8680: I think it's OK to change this test case's prompt template, but I do wonder if the fact that you needed to do so points to a design problem that needs to be addressed. For example, we could choose to gate the lexical-variable tool generation behind a handler so that it's off by default.

eb8680 (commented):
Thanks for simplifying. I took a closer look at the new tests this time and most of them don't seem very useful - they check implementation details that are likely to change or behavior of upstream features that are not part of this new functionality. I also flagged a design question.

effectful/handlers/llm/completions.py
datvo06: Got it. #676 merged so I'll clean this up now.

datvo06:
Update:
1. Make `_LexicalVariableTool` a subtype of `Tool`.
2. Removed the mentioned weak tests and strengthened some to identity rather than equality.

Working on gating LexicalVariableTool behind a handler next as `_LexicalVariableTool` is now a subtype.

datvo06:
Added:
```python
@Operation.define
def expose_lexical_readers() -> bool:
    """Effect controlling whether `_collect_tools` builds synthetic
    read-only Tools for non-Tool/Template values in a Template's
    lexical scope.

    Default behaviour is *off*: only real Tools/Templates/Agents reach
    the LLM, and the lexical context is invisible.  Install
    `LexicalReaders` to flip it on for the call-site where the LLM
    should be able to inspect closure state.
    """
    return False
```
This is determining whether the lexical reader tools are being collected. 
```python
    with handler(LexicalReaders()):
        assert name in _collect_tools(env)
```

effectful/handlers/llm/completions.py
```diff
@@ -178,29 +178,110 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
...
+            f"Reads the value of lexical variable `{name}` from the "
+            f"enclosing scope where this Template was defined.  Takes "
+            f"no arguments; returns the current value."
+        )
+        tool_fn.__annotations__ = {"return": typ}
+        return super().define(tool_fn)
+
+
+@Operation.define
+def expose_lexical_readers() -> bool:
```
eb8680: This is too specific to this use case. Let's promote `_collect_tools` to an `Operation` instead, and put the implementation of the new behavior in a handler instead of in the default rule.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1528,225 @@ def test_tool_forward_ref():
...
+    """With `LexicalReaders` installed, annotated callables, classes,
+    builtins, and methods flow through Encodable's broad Callable
+    handler and become synthesis-shaped tools."""
+    value = make_value()
+    env = {name: value}
+    with handler(LexicalReaders()):
+        assert name in _collect_tools(env)
+
+
+def test_collect_tools_skips_agent_instances_but_exposes_their_tools():
```
eb8680: This does not test functionality in this PR, may change in the future, and should be removed.

tests/test_handlers_llm_template.py
```diff
@@ -1531,3 +1528,225 @@ def test_tool_forward_ref():
...
+        """Doc."""
+        raise NotHandled
+
+    with handler(LexicalReaders()):
+        tools = t.tools
+        assert "_test_data" in tools
+        assert tools["_test_data"]() == [10, 20, 30]
+
+
+def test_lexical_readers_off_by_default():
```
eb8680: This is not testing functionality in this PR and should be removed.

eb8680 (approved):

PR #675, opened 2026-06-08
Keep _-prefixed module imports referenced by emitted stubs (#674)

### LINKED ISSUE #674: Synthesizer drops _-prefixed module imports even when referenced by context type stubs
## Failing test that should pass

A Template that synthesises a callable, called from a pytest test method, fails mypy validation in the synthesizer because the `request` fixture's type (`_pytest.fixtures.TopRequest`) is added to the context stubs but `_pytest` is never imported.

```python
import pytest
from typing import Callable
from effectful.handlers.llm import Template
from effectful.handlers.llm.evaluation import UnsafeEvalProvider


class TestSynthesisRegression:
    @pytest.fixture(autouse=False)
    def _dummy(self): yield

    def test_synthesis_with_pytest_request_in_scope(self, request):
        threshold = 0.85

        @Template.define
        def make() -> Callable[[float], bool]:
            """Use the `threshold` reader, then `def above(x: float) -> bool: return x > <value>`."""
            raise NotImplementedError

        with handler(UnsafeEvalProvider()):
            fn = make()           # currently raises ResultDecodingError

        assert fn(0.9) is True
```

The LLM call itself succeeds and emits `def above(x: float) -> bool: return x > 0.85`. The failure surfaces inside `mypy_type_check`.

## Root cause

`effectful/handlers/llm/evaluation.py:251 collect_imports`:

```python
modules: set[tuple[str, str | None]] = set(
    (k, None)
    for k in sys.modules.keys()
    if k not in SKIPPED_GLOBALS and not k.startswith("_") and k[0].isalpha()
)
```

The `not k.startswith("_")` clause filters out every module whose name begins with an underscore — including `_pytest`, which `pytest`'s public surface re-exports types from. When `collect_variable_declarations` later emits a stub line like

```python
request: _pytest.fixtures.TopRequest
```

(because the runtime `request` value's class is `_pytest.fixtures.TopRequest`), the qualified type name still references `_pytest`, but the import has been silently dropped. mypy then reports

```
<string>:17: error: Name "_pytest" is not defined  [name-defined]
```

and `mypy_type_check` raises `TypeError`, which surfaces as `ResultDecodingError`.

datvo06:
Close #674.

## Fix

Drop the `_`-prefix filter in `collect_imports`. `_pytest.fixtures` and friends now pass through so `mypy_type_check` can resolve ctx stubs like `request: _pytest.fixtures.TopRequest`.

AST-based cleanup might come later (would only import the underscore modules actually referenced by emitted stubs), but it's not blocking and not on the table here.

## Tests

Added corresponding tests.

### DIFF
--- a/effectful/handlers/llm/evaluation.py
+++ b/effectful/handlers/llm/evaluation.py
@@ -248,65 +248,24 @@ SKIPPED_GLOBALS = (
 )
 
 
-def _module_qualnames(nodes: typing.Iterable[ast.AST]) -> set[str]:
-    """Collect every ``a.b.c`` qualified-name *prefix* reachable from
-    each AST node.  Used by ``mypy_type_check`` so that ``collect_imports``
-    keeps ``_``-prefixed modules that the emitted stubs actually
-    reference (e.g. ``_pytest.fixtures.TopRequest``).
-
-    For ``Attribute(Attribute(Name("_pytest"), "fixtures"), "TopRequest")``
-    this returns ``{"_pytest.fixtures"}``.  ``mypy_type_check`` then walks
-    parents to add ``"_pytest"`` as well.
-    """
-    out: set[str] = set()
-    for node in nodes:
-        for sub in ast.walk(node):
-            if not isinstance(sub, ast.Attribute):
-                continue
-            parts: list[str] = [sub.attr]
-            cur: ast.AST = sub.value
-            while isinstance(cur, ast.Attribute):
-                parts.append(cur.attr)
-                cur = cur.value
-            if isinstance(cur, ast.Name):
-                parts.append(cur.id)
-                # parts is now [last_attr, ..., first_name]; the module
-                # qualified name is everything except the final attribute,
-                # reversed.  For `_pytest.fixtures.TopRequest`:
-                # parts == ["TopRequest", "fixtures", "_pytest"]; the
-                # module is "_pytest.fixtures".
-                out.add(".".join(reversed(parts[1:])))
-    return out
-
-
-def collect_imports(
-    ctx: Mapping[str, Any],
-    referenced_module_names: typing.AbstractSet[str] = frozenset(),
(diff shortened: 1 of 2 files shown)


effectful/handlers/llm/evaluation.py
```diff
@@ -248,19 +248,65 @@ def type_to_ast(typ: Any) -> ast.expr:
...
     """
     # (module_name, asname_in_context) for plain imports; asname is None when same as module_name
     modules: set[tuple[str, str | None]] = set(
         (k, None)
         for k in sys.modules.keys()
-        if k not in SKIPPED_GLOBALS and not k.startswith("_") and k[0].isalpha()
+        if k not in SKIPPED_GLOBALS
+        and (
+            # Normal case: public module name (alpha-first, not `_`-prefixed)
+            (k[0].isalpha() and not k.startswith("_"))
```
eb8680: Wouldn't it be easier to just remove this check?

eb8680 (commented):
I think the offending heuristic was originally introduced to simplify the input given to the LLM when a type error is found, but it has soundness issues. Rather than treating these two goals as in conflict, it's probably easier to address them in separate stages - we should generate something we know is correct for the type-checker, and then subsequently clean it up if necessary for presentation to the LLM.

effectful/handlers/llm/evaluation.py
datvo06: It is! It would just make the generated source code long. But as you said, we can move the cleaning up to a later PR.

eb8680: We can address it in another PR, but just to be clear, I meant stages of processing, not just separate PRs - we generate some correct-by-construction input to the type-checker by removing this heuristic, then if that finds a type error we can post-process the code in the error message to remove unhelpful imports.

eb8680 (approved):

PR #676, opened 2026-06-08
Widen nested_type Callable branch on introspection failure (#673)

### LINKED ISSUE #673: nested_type Callable-branch is brittle on values lacking __qualname__/__module__
## Failing tests that should pass

```python
import pytest
from effectful.internals.unification import nested_type


def test_nested_type_marker_decorator_widens_to_class():
    """MarkDecorator lacks `__qualname__`; `typing.get_overloads` raises
    AttributeError instead of widening to `Box(MarkDecorator)`."""
    assert nested_type(pytest.mark.parametrize).value is type(pytest.mark.parametrize)


def test_nested_type_method_descriptor_widens_to_class():
    """method_descriptor lacks `__module__`; same shape."""
    assert nested_type(dict.get).value is type(dict.get)


def test_nested_type_unresolvable_forward_ref_widens():
    """An Operation whose default carries a stringified annotation that
    does not resolve in scope. `inspect.signature` → `__signature__` →
    `typing.get_type_hints` raises NameError instead of widening to
    `Box(Operation)`."""

    def _hidden_module():
        class ClientSession:
            pass

        def real(x: "ClientSession") -> int:
            ...

        return real

    op = defop(_hidden_module())
    assert nested_type(op).value is type(op)
```

## Invariant being violated

IIUC, [and this invariant PR #613,](https://github.com/BasisResearch/effectful/pull/613#issuecomment-2914569567): `nested_type` is part of the canonicalize-widening family: When introspection cannot construct a more precise `TypeExpression`, the documented fallback is `Box(type(value))`, not a raise. The Callable branch in `effectful/internals/unification.py` currently honors this for `ValueError` but lets `AttributeError`/`NameError`/`TypeError` from `typing.get_overloads` and `inspect.signature` escape.

This affected downstream #670 in trying to create reader tools for similar cases to the above.

datvo06:
Close #673.

## Fix

`nested_type` Callable branch was raising rather than widening to `Box(type(value))` on three real values:

- Widen on `AttributeError` / `TypeError` from `typing.get_overloads(value)` — covers `pytest.mark.parametrize` (`MarkDecorator` lacks `__qualname__`) and `dict.get` (`method_descriptor` lacks `__module__`).
- Widen on `NameError` / `AttributeError` / `TypeError` from `inspect.signature(value)` — covers `Operation` whose `__signature__` runs `typing.get_type_hints` and hits an unresolvable forward-ref.
- Propagate the widening through the `Operation` branch when its Callable delegate returns a bare type (`typing.get_args` was failing to unpack).

Per the canonicalize-widening principle on #613.

## Tests

Added corresponding tests.

### DIFF
--- a/tests/test_internals_unification.py
+++ b/tests/test_internals_unification.py
@@ -867,6 +867,73 @@ def test_nested_type_term_error():
         nested_type(mock_term)
 
 
+def test_nested_type_marker_decorator_widens_to_class():
+    """#673: ``pytest.mark.parametrize`` is a ``MarkDecorator`` instance --
+    callable, but lacks ``__qualname__``. The Callable branch's
+    ``typing.get_overloads`` raises ``AttributeError``; per the
+    canonicalize-widening principle (PR #613) the fallback should be
+    ``Box(type(value))``."""
+    val = pytest.mark.parametrize
+    assert nested_type(val).value is type(val)
+
+
+def test_nested_type_method_descriptor_widens_to_class():
+    """#673: ``dict.get`` is a ``method_descriptor`` -- callable but
+    missing ``__module__``. Same widening shape as MarkDecorator."""
+    val = dict.get
+    assert nested_type(val).value is type(val)
+
+
+def test_nested_type_unresolvable_forward_ref_widens():
+    """#673: an ``Operation`` whose default carries a stringified
+    annotation whose name does not resolve in the captured scope.
+    ``inspect.signature`` consults ``Operation.__signature__`` which
+    calls ``typing.get_type_hints`` -- forward-ref evaluation raises
+    ``NameError``. ``nested_type`` should widen to ``Box(type(value))``
+    rather than propagating."""
+    from effectful.ops.syntax import defop
+
+    def _hidden_module():
+        class ClientSession:  # noqa: F841 -- intentionally not visible to op
+            pass
+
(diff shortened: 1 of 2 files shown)


tests/test_internals_unification.py
```diff
@@ -867,6 +867,45 @@ def test_nested_type_term_error():
...
+    calls ``typing.get_type_hints`` -- forward-ref evaluation raises
+    ``NameError``. ``nested_type`` should widen to ``Box(type(value))``
+    rather than propagating."""
+    from effectful.ops.syntax import defop
+
+    def _hidden_module():
+        class ClientSession:  # noqa: F841 -- intentionally not visible to op
+            pass
+
+        def real(x: "ClientSession") -> int:  # noqa: F821 -- forward ref unresolved
```
eb8680, line 900: What about the case when `ClientSession` is not quoted?

eb8680 (commented):
This is missing the eager-annotation case from #673

tests/test_internals_unification.py
datvo06: I just added `test_nested_type_eager_annotation_produces_precise_type` in 2b8c274. When the annotation is unquoted it resolves at function-def time, so `nested_type` returns the precise `Operation[[ClientSession], int]`

eb8680 (approved):

PR #679, opened 2026-06-10
Add unify rule for Operation with Callable

### LINKED ISSUE #669: `Operation` should unify with `Callable`
The following should work:

```
from typing import Callable

from effectful.ops.types import NotHandled, Operation


@Operation.define
def f(x: int) -> int:
    raise NotHandled


@Operation.define
def g[S, T](x: Callable[[S], T]) -> T:
    raise NotHandled


t = g(f)
```

eb8680:
Fixes #669 

This PR adds a case in `unify` to handle unifying `Operation` and `Callable`, and a few tests including a regression test based on the example in the original issue.

### DIFF
--- a/effectful/internals/unification.py
+++ b/effectful/internals/unification.py
@@ -539,6 +539,17 @@ def _unify_generic(typ, subtyp, subs: Substitutions) -> Substitutions:
             typing.get_origin(typ), collections.abc.Generator
         ):
             return unify(typing.get_args(typ)[0], typing.get_args(subtyp)[0], subs)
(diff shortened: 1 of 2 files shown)


effectful/internals/unification.py
```diff
@@ -556,6 +567,17 @@ def _unify_generic(typ, subtyp, subs: Substitutions) -> Substitutions:
         and issubclass(subtyp, typing.get_origin(typ))
     ):
         return subs  # implicit expansion to subtyp[Any]
+    elif isinstance(typ, GenericAlias):
```
eb8680, line 570: This is from @jfeser in #656

jfeser (approved):

Issue #685, opened 2026-06-11
RestrictedEvalProvider.exec drops rebindings of seeded names and does not wire RestrictedPython's print protocol


datvo06:
The following code should pass with `RestrictedEvalProvider`:

```python
with handler(RestrictedEvalProvider()):
    ns = {"x": 1}
    exec(compile(parse("x = 99", "<f>"), "<f>"), ns)
    assert ns["x"] == 99      # FAILS: rebind of seeded name dropped, ns["x"] == 1

    exec(compile(parse("print('hi')", "<f>"), "<f>"), ns)
    # FAILS: NameError: name '_print_' is not defined
```

Cause of the first: `keys_before` is snapshotted *after* `rglobals.update(env)`, so the copy-back only sees never-before-present keys. Cause of the second: RestrictedPython rewrites `print(...)` to its `_print_` collector protocol and the provider never supplies `_print_`.

Fine for single-shot synthesis decoding (the namespace is discarded after one read), but breaks any caller that reuses the namespace across calls — blocks #678, whose v1 is fenced to `UnsafeEvalProvider` because of this. Related: #573, #490.

datvo06:
Thinking of fixing this as follow, in `RestrictedEvalProvider.exec`:

```python
before = dict(rglobals)                      # snapshot AFTER seeding
builtins.exec(bytecode, rglobals, rglobals)
sentinel = object()
for k, v in rglobals.items():
    if k != "__builtins__" and before.get(k, sentinel) is not v:
        env[k] = v
```

PR #687, opened 2026-06-11
Add persistent Python REPL tool for Templates

### LINKED ISSUE #678: Consider adding generic code execution tools
We should add generic code execution tools to `handlers.llm.evaluation` that `Template`s can use prior to computing a final answer.

One nice way to do this could be to extend the [`InteractiveConsole`](https://docs.python.org/3/library/code.html#code.InteractiveConsole) class in the [`code`](https://docs.python.org/3/library/code.html) standard library module, which is a substrate for implementing customized REPL-like interfaces to the Python language runtime. This would make it easier to build a toolset that is sensitive to `Template` lexical context, has persistent state beyond a single tool call and allows redirecting output streams into LLM context.

datvo06:
Closes #678.

Adds a generic code-execution tool a Template's LLM can call to run Python before producing a final answer. Few desiderata:
(1) linked to the Template's lexical context
(2) persistent state across tool calls
(3) redirected output streams.

**Example.**

```python
from effectful.handlers.llm import Template, LiteLLMProvider
from effectful.handlers.llm.completions import PythonRepl
from effectful.handlers.llm.evaluation import UnsafeEvalProvider

readings = [12, 19, 23, 31, 8, 27]

@Template.define
def outlier_count() -> int:
    """Use the `exec_code` tool to compute how many values in `readings`
    lie more than one population stdev from the mean; return that count."""
    raise NotHandled

with handler(LiteLLMProvider()), handler(UnsafeEvalProvider()), handler(PythonRepl()):
    n = outlier_count()
```

The LLM runs code across rounds with state persisting:

```
exec_code("m = ...; s = ...")   ->  ""        # readings came from lexical scope
exec_code("print(...using m, s...)")  ->  "2\n"   # m, s persisted from the prior call
```

**`ReplSession`** (evaluation.py): a plain class seeded from a lexical context. Each `run(source)` executes one complete snippet in exec mode through the `parse`/`compile`/`exec` effect operations (so the installed eval-provider owns sandboxing), captures stdout/stderr into a per-call buffer, and persists bindings in `self.locals` across calls. Per-snippet filenames keep each cell's source in `linecache` so cross-snippet tracebacks format correctly; the traceback formatter trims the effect-machinery frames so the LLM sees only its own code.

**`PythonRepl`** (completions.py): a `collect_tools` handler exposing an `exec_code` Tool bound to a session that persists for the Template invocation. Off by default. Sessions are keyed by `id(env)` and pruned via `weakref.finalize` (no leak, no id-reuse), created lazily on first use.

The `exec` op docstring now states its binding-effects contract: after `exec(bytecode, env)`, `env` reflects all top-level bindings — new and rebound alike.

**Fenced to `UnsafeEvalProvider` (#685).** `RestrictedEvalProvider` drops rebindings of seeded names and lacks RestrictedPython's print wiring. The two restricted REPL tests are `xfail(strict)` against #685 and flip to xpass when it lands.

**Tests.** 12 `ReplSession` laws and 7 `PythonRepl` handler tests, deterministic under `UnsafeEvalProvider` (seed, persistence, rebind, multi-statement, print, syntax-error, exception isolation, KeyboardInterrupt-propagation, traceback trim, cross-snippet traceback, reentrancy, effect-routing; off-by-default, exposes-tool, name-collision, composes-with-LexicalReaders, lazy creation, same-session-across-rounds, distinct-env). Plus one replayed gpt-4o-mini integration test where the model genuinely uses `exec_code` across rounds to compute a statistic.

Decoupled from `LexicalReaders` by default; coupling REPL-created symbols into the readers is noted as a follow-up.

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -478,9 +478,12 @@ def call_tool(tool_call: DecodedToolCall) -> Message:
             *tool_call.bound_args.args, **tool_call.bound_args.kwargs
         )
     except ReplExecutionError as e:
-        # `exec_code`'s snippet raised: surface its trimmed transcript verbatim.
+        # `exec_code`'s snippet raised: surface its trimmed transcript verbatim,
+        # and expose the *real* exception so `catch_tool_errors` can match on it.
         raise CodeExecutionError(
-            raw_tool_call=tool_call, original_error=e, transcript=e.transcript
+            raw_tool_call=tool_call,
+            original_error=e.original_error,
+            transcript=e.transcript,
         ) from e
     except Exception as e:
         raise ToolCallExecutionError(raw_tool_call=tool_call, original_error=e) from e
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -31,7 +31,7 @@ from effectful.handlers.llm.encoding import (
     Encodable,
     to_content_blocks,
 )
-from effectful.handlers.llm.evaluation import ReplSession
+from effectful.handlers.llm.evaluation import ReplExecutionError, ReplSession
 from effectful.handlers.llm.template import (
     Agent,
     Template,
@@ -173,6 +173,27 @@ class ToolCallExecutionError[E: Exception, T](DecodingError[E]):
                 "role": "tool",
                 "tool_call_id": self.raw_tool_call.id,
                 "content": error_message,
+            },
+        )
+
+
+@dataclasses.dataclass
+class CodeExecutionError(ToolCallExecutionError):
+    """An `exec_code` snippet raised at runtime.
+
+    `ReplSession` already builds a self-contained, frame-trimmed transcript, so
+    the feedback is that ``transcript`` (captured stdout/stderr plus the user's
+    own traceback) verbatim rather than a generic stack trace.
+    """
+
+    transcript: str = ""
+
+    def to_feedback_message(self, include_traceback: bool) -> Message:
+        return _make_message(
+            {
+                "role": "tool",
+                "tool_call_id": self.raw_tool_call.id,
+                "content": self.transcript,
             },
         )
 
@@ -456,6 +477,11 @@ def call_tool(tool_call: DecodedToolCall) -> Message:
         result = tool_call.tool(
             *tool_call.bound_args.args, **tool_call.bound_args.kwargs
         )
+    except ReplExecutionError as e:
+        # `exec_code`'s snippet raised: surface its trimmed transcript verbatim.
+        raise CodeExecutionError(
+            raw_tool_call=tool_call, original_error=e, transcript=e.transcript
+        ) from e
     except Exception as e:
         raise ToolCallExecutionError(raw_tool_call=tool_call, original_error=e) from e
 
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -10,6 +10,7 @@ import textwrap
 import traceback
 import typing
 import uuid
+import weakref
 
 import litellm
 import pydantic
@@ -30,6 +31,7 @@ from effectful.handlers.llm.encoding import (
     Encodable,
     to_content_blocks,
 )
+from effectful.handlers.llm.evaluation import ReplSession
 from effectful.handlers.llm.template import (
     Agent,
     Template,
@@ -287,6 +289,65 @@ class LexicalReaders(ObjectInterpretation):
             ):
                 continue
         return result
+
+
+class PythonRepl(ObjectInterpretation):
+    """Override `collect_tools` to expose a persistent Python session as an
+    `exec_code` Tool.
+
+    Off by default; install it where the LLM should be able to run code whose
+    state (variables, imports, definitions) survives across tool calls within
+    one Template invocation.  The session is keyed by the Template's lexical
+    context, seeded from it, and routes execution through the
+    `parse`/`compile`/`exec` effect operations.
+
+    v1 requires `UnsafeEvalProvider`: `RestrictedEvalProvider`'s exec drops
+    rebindings of seeded names and does not wire RestrictedPython's print
+    collector (tracked in #685), both of which break the REPL contract.
+    """
+
+    def __init__(self):
+        self._sessions: dict[int, ReplSession] = {}
+
+    def _session_for(
+        self, env: collections.abc.Mapping[str, typing.Any]
+    ) -> ReplSession:
+        key = id(env)
+        session = self._sessions.get(key)
+        if session is None:
+            # Prune the entry the instant `env` is collected: no leak, and the
+            # stale id can never alias a later object.  `pop` is bound to the
+            # dict (not `env`), so the finalizer does not keep `env` alive.
+            # `env` must be weak-referenceable -- the driver always passes a
+            # ChainMap (which is); guard the misuse path (e.g. a plain dict)
+            # with a clear message rather than skipping the finalizer, which
+            # would reintroduce the id-reuse hazard.
+            if not hasattr(env, "__weakref__"):
+                raise TypeError(
+                    f"PythonRepl needs a weak-referenceable lexical context to key "
+                    f"its session; got {type(env).__name__}, which is not. "
+                    f"Templates pass a ChainMap; only direct misuse hits this."
+                )
+            session = ReplSession(env)
+            self._sessions[key] = session
+            weakref.finalize(env, self._sessions.pop, key, None)
+        return session
+
+    @implements(collect_tools)
+    def _collect(
+        self, env: collections.abc.Mapping[str, typing.Any]
+    ) -> collections.abc.Mapping[str, Tool]:
+        tools = dict(fwd())
+
+        @Tool.define
+        def exec_code(source: str) -> str:
+            """Execute Python in a persistent session.  Variables, imports and
+            definitions carry across calls.  Returns the captured stdout and
+            stderr (use print() to inspect values)."""
+            return self._session_for(env).run(source)
+
+        tools.setdefault("exec_code", exec_code)  # a real user tool of same name wins
+        return tools
 
 
 @Operation.define
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -291,7 +291,7 @@ class LexicalReaders(ObjectInterpretation):
 
 
 @Operation.define
-def _repl_session(env: collections.abc.Mapping[str, typing.Any]) -> ReplSession:
+def _repl_session(env: collections.abc.MutableMapping[str, typing.Any]) -> ReplSession:
     """Return the REPL session for the current Template call, seeded from `env`.
 
     `PythonRepl` installs a fresh handler for this inside each `Template.__apply__`
@@ -334,7 +334,7 @@ class PythonRepl(ObjectInterpretation):
         session: ReplSession | None = None
 
         def session_for(
-            env: collections.abc.Mapping[str, typing.Any],
+            env: collections.abc.MutableMapping[str, typing.Any],
         ) -> ReplSession:
             nonlocal session
             if session is None:
@@ -349,7 +349,12 @@ class PythonRepl(ObjectInterpretation):
         self, env: collections.abc.Mapping[str, typing.Any]
     ) -> collections.abc.Mapping[str, Tool]:
         tools = dict(fwd())
-        tools["exec_code"] = _repl_session(env).exec_code
+        # `collect_tools` only promises a `Mapping`, but the per-call `env` is the
+        # writable `ChainMap` the session splices its shared scope layer into, so
+        # narrow it for `_repl_session`/`ReplSession`.
+        tools["exec_code"] = _repl_session(
+            typing.cast(collections.abc.MutableMapping[str, typing.Any], env)
+        ).exec_code
         return tools
 
 
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -349,11 +349,7 @@ class PythonRepl(ObjectInterpretation):
         self, env: collections.abc.Mapping[str, typing.Any]
     ) -> collections.abc.Mapping[str, Tool]:
         tools = dict(fwd())
-        # Reuse the session the enclosing `Template.__apply__` already installed
-        # for this call (see `_repl_session`) and expose its `exec_code` method
-        # directly -- the tool's behaviour, docstring and name all come from
-        # `ReplSession.exec_code`, nothing is hidden in this handler.
-        tools["exec_code"] = Tool.define(_repl_session(env).exec_code)
+        tools["exec_code"] = _repl_session(env).exec_code
         return tools
 
 
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -315,21 +315,13 @@ class PythonRepl(ObjectInterpretation):
         key = id(env)
         session = self._sessions.get(key)
         if session is None:
+            session = ReplSession(env)
+            self._sessions[key] = session
             # Prune the entry the instant `env` is collected: no leak, and the
             # stale id can never alias a later object.  `pop` is bound to the
             # dict (not `env`), so the finalizer does not keep `env` alive.
-            # `env` must be weak-referenceable -- the driver always passes a
-            # ChainMap (which is); guard the misuse path (e.g. a plain dict)
-            # with a clear message rather than skipping the finalizer, which
-            # would reintroduce the id-reuse hazard.
-            if not hasattr(env, "__weakref__"):
-                raise TypeError(
-                    f"PythonRepl needs a weak-referenceable lexical context to key "
-                    f"its session; got {type(env).__name__}, which is not. "
-                    f"Templates pass a ChainMap; only direct misuse hits this."
-                )
-            session = ReplSession(env)
-            self._sessions[key] = session
+            # `env` is the Template's lexical context (a ChainMap); a non
+            # weak-referenceable env (e.g. a bare dict) raises here, loudly.
             weakref.finalize(env, self._sessions.pop, key, None)
         return session
 
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -475,9 +475,6 @@ def call_tool(tool_call: DecodedToolCall) -> Message:
             *tool_call.bound_args.args, **tool_call.bound_args.kwargs
         )
     except Exception as e:
-        # An `exec_code` snippet error arrives here as a `ReplExecutionError`
-        # whose ``str`` is the trimmed transcript, so the generic feedback path
-        # surfaces it -- no bespoke exception type needed.
         raise ToolCallExecutionError(raw_tool_call=tool_call, original_error=e) from e
 
     return_type: pydantic.TypeAdapter[typing.Any] = pydantic.TypeAdapter(
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -10,7 +10,6 @@ import textwrap
 import traceback
 import typing
 import uuid
-import weakref
 
 import litellm
 import pydantic
@@ -31,7 +30,7 @@ from effectful.handlers.llm.encoding import (
     Encodable,
     to_content_blocks,
 )
-from effectful.handlers.llm.evaluation import ReplExecutionError, ReplSession
+from effectful.handlers.llm.evaluation import ReplSession
 from effectful.handlers.llm.template import (
     Agent,
     Template,
@@ -177,27 +176,6 @@ class ToolCallExecutionError[E: Exception, T](DecodingError[E]):
         )
 
 
-@dataclasses.dataclass
-class CodeExecutionError(ToolCallExecutionError):
-    """An `exec_code` snippet raised at runtime.
-
-    `ReplSession` already builds a self-contained, frame-trimmed transcript, so
-    the feedback is that ``transcript`` (captured stdout/stderr plus the user's
-    own traceback) verbatim rather than a generic stack trace.
-    """
-
-    transcript: str = ""
-
-    def to_feedback_message(self, include_traceback: bool) -> Message:
-        return _make_message(
-            {
-                "role": "tool",
-                "tool_call_id": self.raw_tool_call.id,
-                "content": self.transcript,
-            },
-        )
-
-
 type MessageResult[T] = tuple[Message, typing.Sequence[DecodedToolCall], T | None]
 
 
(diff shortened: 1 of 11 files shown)


datvo06:
This doesn't close #685 yet, as it is right now. I'm still deciding how persistent the state between different call this really is (shared between one agent, or all agents and all templates). Also there might be a new problem with OpenAI trimming headers. Investigating.

eb8680:
> how persistent the state between different call this really is (shared between one agent, or all agents and all templates)

I think you only want it to persist for the duration of one `Template` call, not to share across multiple calls or multiple agents.

datvo06:
I see, that is simplifying! On it now.

datvo06:
Another question that's coming up is if there are nested template:
```python
@Template.define
def foo():
      """Some prompt"""
      raise NotHandled


@Template.define
def bar():
      """Some prompt"""
      raise NotHandled
```
Then will the lexical context/envs be shared between `foo` and `bar` if `bar()` calls `exec()` and then `foo()` as tools? My current take is that it will be separated, `bar()` calling `exec()` will have no effect on `foo()` calling `exec()` down in the call stack.

eb8680:
Code should be understood to live in the lexical context of the relevant `Template` body. The body of `bar` can read the shared lexical context across `foo`, `bar` and whatever else is in the same scope, as well as any arguments to `bar` (none in this example), but any new variables are local to the body of `bar` and hence invisible to `foo`.

datvo06:
So far, this is what this PR did:
1. Added `ReplSession` based on `code.InteractiveInterpreter`. This add an indirection between `InteractiveInterpreter` and `exec_code` tool, allow to use existing sandboxing for `EvalHandlers`. 
2. Add `ReplExecutionError`, so that we can later use it jointly with `RetryLLMHandler`. 
3. Add `PythonRepl` handler that exposes `exec_code` tool, also maintain ids between Template calls so that each Template gets its own session.
4. Added tests (also retry tests)

effectful/handlers/llm/completions.py
```diff
@@ -175,6 +177,27 @@ def to_feedback_message(self, include_traceback: bool) -> Message:
         )
 
 
+@dataclasses.dataclass
+class CodeExecutionError(ToolCallExecutionError):
```
eb8680: This seems like an unnecessary layer of indirection. Just raise an error with the right content in the code execution tool, so you don't need this at all.

effectful/handlers/llm/completions.py
```diff
@@ -289,6 +312,57 @@ def _collect(
...
+
+    v1 requires `UnsafeEvalProvider`: `RestrictedEvalProvider`'s exec drops
+    rebindings of seeded names and does not wire RestrictedPython's print
+    collector (tracked in #685), both of which break the REPL contract.
+    """
+
+    def __init__(self):
+        self._sessions: dict[int, ReplSession] = {}
+
+    def _session_for(
```
eb8680: Needing to use `weakref` is almost always a code smell that points to design flaws. In this case, we shouldn't carry around multiple `ReplSession` instances in one global handler instance in this way.

effectful/handlers/llm/completions.py
```diff
@@ -403,6 +477,14 @@ def call_tool(tool_call: DecodedToolCall) -> Message:
         result = tool_call.tool(
             *tool_call.bound_args.args, **tool_call.bound_args.kwargs
         )
+    except ReplExecutionError as e:
```
eb8680: As mentioned above, this should not be necessary if the tool raises an appropriate error.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +817,122 @@ def exec(
...
+    is no bare-expression auto-echo, so use `print()` to surface values.
+    """
+
+    def __init__(self, env: Mapping[str, Any]):
+        super().__init__(dict(env))
+        self._counter = itertools.count()
+        # Parallel stacks so re-entrant `run` calls (exec'd code that calls back
+        # into the same session) each keep their own transcript and the
+        # exception (if any) the run raised.
+        self._buffers: list[io.StringIO] = []
```
eb8680: Making `ReplSession` carry state about multiple sessions is clunky and overcomplicated. A `ReplSession` instance should only be responsible for one session, then none of these state variable lists would be necessary.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +817,122 @@ def exec(
         for key in rglobals:
             if key not in keys_before:
                 env[key] = rglobals[key]
+
+
+_REPL_FILENAME_PREFIX = "<exec_code-"
```
eb8680: nit: make this a `typing.ClassVar` on `ReplSession` rather than a module global.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +817,122 @@ def exec(
...
+        # Parallel stacks so re-entrant `run` calls (exec'd code that calls back
+        # into the same session) each keep their own transcript and the
+        # exception (if any) the run raised.
+        self._buffers: list[io.StringIO] = []
+        self._error: list[Exception | None] = []
+        # Replace the native single-mode CommandCompiler with our op-routed,
+        # linecache-populating compiler.
+        self.compile = self._compile_through_ops  # type: ignore[assignment]
+
+    def _compile_through_ops(self, source: str, filename: str, symbol: str) -> CodeType:
```
eb8680: nit: put this in a separate subclass of `codeop.CommandCompiler`, mirroring the implementation of `InteractiveInterpreter` more closely.

effectful/handlers/llm/completions.py
```diff
@@ -289,6 +312,57 @@ def _collect(
...
+        return session
+
+    @implements(collect_tools)
+    def _collect(
+        self, env: collections.abc.Mapping[str, typing.Any]
+    ) -> collections.abc.Mapping[str, Tool]:
+        tools = dict(fwd())
+
+        @Tool.define
+        def exec_code(source: str) -> str:
```
eb8680: Rather than being hidden here, `exec_code` should probably be a method of the `ReplSession` class, with a more precise name and a more descriptive docstring.

effectful/handlers/llm/completions.py
```diff
@@ -289,6 +312,57 @@ def _collect(
...
+            """Execute Python in a persistent session.  Variables, imports and
+            definitions carry across calls.  Returns the captured stdout and
+            stderr (use print() to inspect values)."""
+            return self._session_for(env).run(source)
+
+        tools.setdefault("exec_code", exec_code)  # a real user tool of same name wins
```
eb8680: This feels like a land mine that will surely cause impossible-to-diagnose problems for someone at some point down the line. It should be removed. If `exec_code` is too generic, it should get a more precise name. This also suggests considering a more general change of using `__qualname__` to specify tool names rather than just `__name__`, but that's out of scope for this PR.

effectful/handlers/llm/completions.py
```diff
@@ -289,6 +312,57 @@ def _collect(
         return result
 
 
+class PythonRepl(ObjectInterpretation):
```
eb8680, line 306: The right way to implement the behavior implemented by this handler is probably to have it handle `_collect_tools` to inject a code execution tool into the tool set, and introduce a fresh handler for that tool inside every `Template.__apply__`. That way there is no separate global stack of `ReplSession`s; each session is introduced and eliminated by its own handler which is restricted to the duration of the `Template` call by construction.

This is the same pattern by which the `__history__` state variable is managed for `Template` calls.

tests/test_handlers_llm_evaluation.py
```diff
@@ -1537,3 +1540,140 @@ def test_builtins_in_env_does_not_bypass_security():
...
+    """The transcript for an uncaught exception shows only the user's snippet
+    frames, not effectful's internal frames."""
+    with handler(UnsafeEvalProvider()):
+        with pytest.raises(ReplExecutionError) as exc_info:
+            ReplSession({}).run("1 / 0")
+        out = exc_info.value.transcript
+        assert "ZeroDivisionError" in out
+        assert "runtime.py" not in out
+        assert "ops/types.py" not in out
+        assert "exec_code-" in out  # the user frame's per-snippet filename
```
eb8680: All of these names are implementation details that may change, the tests should not be sensitive to them.

I assume most of this test code is AI-generated, which is fine, but I think I've made versions of this same comment repeatedly on several of your recent PRs. A human developer would remember receiving similar feedback and address it before requesting a review on a new PR, so if you want human reviewers to provide fast and high-quality reviews for AI-generated code it's good practice to try to reduce the cognitive burden on them by helping the AI anticipate their feedback, perhaps by creating a custom skill.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +817,122 @@ def exec(
...
+        while user_tb is not None and user_tb.tb_frame.f_globals is not self.locals:
+            user_tb = user_tb.tb_next
+        self.write("".join(traceback.format_exception(exc_type, exc, user_tb or tb)))
+
+    def write(self, data: str) -> None:
+        # Error output from show{traceback,syntaxerror} lands here; route it to
+        # the current call's buffer (top of stack) for re-entrancy safety.
+        self._buffers[-1].write(data)
+
+    def run(self, source: str) -> str:
```
eb8680: This should probably be the `Tool` that's exposed to the LLM.

eb8680 (commented):
For the execution tool, I wonder if we should make it take a compiled `types.CodeType` as an argument rather than a source code string. If we did that and implemented `Encodable` for `types.CodeType`, we might be able to catch and resolve syntax and compilation errors early, like we do for type errors in other tool call arguments.

I also wonder if its return type should distinguish between `stderr` and `stdout` by returning a pair of strings rather than just one. Maybe that's overcomplicating things for the LLM and hence not worth doing for now.

tests/test_handlers_llm_evaluation.py
datvo06: Thanks! Sorry, I missed this. I will create a better skill.

effectful/handlers/llm/completions.py
```diff
@@ -404,6 +475,9 @@ def call_tool(tool_call: DecodedToolCall) -> Message:
             *tool_call.bound_args.args, **tool_call.bound_args.kwargs
         )
     except Exception as e:
+        # An `exec_code` snippet error arrives here as a `ReplExecutionError`
```
eb8680: nit: remove this comment - this is an LLM explaining a diff to you, not something that's useful outside the context of this PR.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+        REPL.  There is no bare-expression auto-echo, so use print() to surface
+        values.  A snippet that raises propagates the error (carrying its
+        transcript) rather than returning, so it behaves like a normal Python
+        call.
+        """
+        # Per-snippet filename so each cell's source is retained in `linecache`
+        # independently -- a single shared name would overwrite earlier cells
+        # and format earlier-defined functions' tracebacks against the wrong
+        # source.
+        filename = f"{self._FILENAME_PREFIX}{next(self._counter)}>"
```
eb8680: This will generate collisions across `Template` calls, get rid of `self._counter` and use `uuid.uuid4` or similar instead.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+        exc_type, exc, tb = sys.exc_info()
+        user_tb = tb
+        while user_tb is not None and user_tb.tb_frame.f_globals is not self.locals:
+            user_tb = user_tb.tb_next
+        self.write("".join(traceback.format_exception(exc_type, exc, user_tb or tb)))
+
+    def write(self, data: str) -> None:
+        # Error output from show{traceback,syntaxerror} lands here; route it to
+        # the current run's buffer.
+        self._buffer.write(data)
```
eb8680: Can this just defer to `super().write`?
```py
with contextlib.redirect_stdout(self._buffer), contextlib.redirect_stderr(self._buffer):
  return super().write(data)
```

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+        values.  A snippet that raises propagates the error (carrying its
+        transcript) rather than returning, so it behaves like a normal Python
+        call.
+        """
+        # Per-snippet filename so each cell's source is retained in `linecache`
+        # independently -- a single shared name would overwrite earlier cells
+        # and format earlier-defined functions' tracebacks against the wrong
+        # source.
+        filename = f"{self._FILENAME_PREFIX}{next(self._counter)}>"
+        self._buffer = io.StringIO()
```
eb8680: Why is `self._buffer` being overwritten in each `exec_code` call? Shouldn't it persist across every call in the instance?

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+                contextlib.redirect_stderr(self._buffer),
+            ):
+                self.runsource(source, filename)
+        finally:
+            transcript = self._buffer.getvalue()
+            error = self._error
+        if error is not None:
+            # Like a normal Python call, a failed snippet raises rather than
+            # returning; `call_tool` catches it and feeds the transcript back.
+            raise ReplExecutionError(transcript, error)
```
eb8680: Thinking more about this, I'm not sure `exec_code` needs to `raise` at all, as long as any error/traceback generated by `exec` is included in the transcript that's returned at the end. That also means there's no need to track and store the exception object.

This simplification should not lead to meaningfully different behavior from the current implementation; `RetryLLMHandler` just appends runtime error messages from tool calls and continues the assistant loop with the extra context.

Errors generated by staticaclly malformed tool arguments are handled differently, before the tool function is called at all, so no version of this logic would take advantage of that - the only way to do so would be to use a different, stronger type for the argument(s) to `exec_code`.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+
+    def runcode(self, code_obj: CodeType) -> None:
+        try:
+            exec(code_obj, self.locals)
+        except Exception as exc:
+            # `except Exception` lets SystemExit/KeyboardInterrupt propagate.
+            self.showtraceback()
+            self._error = exc
+
+    def showsyntaxerror(self, *args: Any, **kwargs: Any) -> None:
```
eb8680: I think this can be removed, we can just use the underlying `InteractiveInterpreter.showsyntaxerror()` with no changes.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+        # A syntax/compile error is a failed run too; capture it the way
+        # `runcode` captures a runtime error, so `exec_code` raises for both.  The
+        # SyntaxError is the in-flight exception during `runsource`'s except.
+        exc = sys.exc_info()[1]
+        assert isinstance(exc, Exception)
+        self._error = exc
+        super().showsyntaxerror(*args, **kwargs)
+
+    def showtraceback(self) -> None:
```
eb8680: Can we just use `InteractiveInterpreter.showtraceback()` and remove this? I think there is some extra behavior there that this is bypassing, and I also don't really trust this frame-dropping logic.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+    _FILENAME_PREFIX: typing.ClassVar[str] = "<exec_code-"
+
+    def __init__(self, env: Mapping[str, Any]):
+        super().__init__(dict(env))
+        self._counter = itertools.count()
+        # One session per instance, so a single in-flight `exec_code` owns these: the
+        # buffer captures that run's stdout/stderr, and `_error` holds the
+        # exception it raised, if any.
+        self._buffer = io.StringIO()
+        self._error: Exception | None = None
```
eb8680: As noted below I don't think we need to track this ourselves, it can be removed.

effectful/handlers/llm/evaluation.py
```diff
@@ -809,3 +818,141 @@ def exec(
...
+        # Replace the native single-mode CommandCompiler with our op-routed,
+        # linecache-populating compiler.
+        self.compile = _OpCommandCompiler()
+
+    def runcode(self, code_obj: CodeType) -> None:
```
eb8680: nit: this should mirror `super().runcode` exactly, including the argument name; the only difference is the redefined `exec` symbol:
```py
    def runcode(self, code: CodeType) -> None:
        try:
            exec(code, self.locals)
        except SystemExit:
            raise
        except:
            self.showtraceback()
```

effectful/handlers/llm/completions.py
```diff
@@ -289,6 +290,76 @@ def _collect(
...
+    @implements(collect_tools)
+    def _collect(
+        self, env: collections.abc.Mapping[str, typing.Any]
+    ) -> collections.abc.Mapping[str, Tool]:
+        tools = dict(fwd())
+        # Reuse the session the enclosing `Template.__apply__` already installed
+        # for this call (see `_repl_session`) and expose its `exec_code` method
+        # directly -- the tool's behaviour, docstring and name all come from
+        # `ReplSession.exec_code`, nothing is hidden in this handler.
+        tools["exec_code"] = Tool.define(_repl_session(env).exec_code)
```
eb8680: Put the `Tool.define` on the definition of `exec_code`, not here. `Tool.__get__` will generate per-instance `Tool`s automatically.

datvo06: Ah, right! If the tool is imported from completion it will appear in lexical scope and tool collector will collect it.

eb8680: No, I just mean that if you decorate the definition of `ReplSession.exec_code` with `Tool.define`:
```py
class ReplSession(...):
    ...
    @Tool.define
    def exec_code(self, ...): ...

...

repl1 = ReplSession()
repl2 = ReplSession()
assert isinstance(repl1.exec_code, Tool)
assert isinstance(repl2.exec_code, Tool)
assert repl1.exec_code is not repl2.exec_code
assert len(inspect.signature(repl1.exec_code).parameters) < len(inspect.signature(ReplSession.exec_code).parameters)
```

effectful/handlers/llm/evaluation.py
datvo06: Got it! Removed

effectful/handlers/llm/evaluation.py
```diff
@@ -91,12 +97,60 @@ def exec(
...
     """
     raise NotImplementedError(
         "An eval provider must be installed in order to execute code."
     )
 
 
+_CODE_FILENAME_PREFIX = "<exec_code-"
+
+
+class EncodableCode(pydantic.BaseModel):
```
eb8680: Rather than being exposed as part of a public API, I would expect `EncodableCode` to live in `encoding.py` and be the  internal `Encodable` implementation for `types.CodeType`, analogous to `handlers.llm.encoding._ComplexModel` or `SynthesizedFunction`. I don't think it carries any extra information that is not available from `types.CodeType` and `linecache`/`inspect.getsource`.

effectful/handlers/llm/evaluation.py
```diff
@@ -829,3 +883,58 @@ def exec(
...
+    `EncodableCode` is built; this session only executes.
+    """
+
+    def __init__(self, env: Mapping[str, Any]):
+        # `InteractiveInterpreter.__init__` sets `self.locals`; subclassing lets
+        # us reuse its `runcode`/`showtraceback`/`write` machinery rather than
+        # re-implementing traceback capture by hand.
+        super().__init__(dict(env))
+        # The session's accumulated stdout/stderr, persisting for its lifetime.
+        self._buffer = io.StringIO()
```
eb8680: nit: rename to `self.stderr` and add an annotation `stderr: io.StringIO` to the class body, we might want to use this to introspect a `ReplSession`.

effectful/handlers/llm/evaluation.py
```diff
@@ -829,3 +883,58 @@ def exec(
...
+
+    Compilation -- and therefore syntax checking -- happens earlier, when the
+    `EncodableCode` is built; this session only executes.
+    """
+
+    def __init__(self, env: Mapping[str, Any]):
+        # `InteractiveInterpreter.__init__` sets `self.locals`; subclassing lets
+        # us reuse its `runcode`/`showtraceback`/`write` machinery rather than
+        # re-implementing traceback capture by hand.
+        super().__init__(dict(env))
```
eb8680: I think this should pass on `env` without the `dict` shallow copy and require `env` to be a `MutableMapping` or `collections.ChainMap`. This would allow `ReplSession` to affect the rest of the `Template` call, not just itself, mirroring `exec`

In practice, `env` should always be a `ChainMap` where the first mapping contains the arguments of the `Template` call (created [here](https://github.com/BasisResearch/effectful/blob/c318d3c6ccf202c131c16d1be8ccf74dece3f163/effectful/handlers/llm/completions.py#L596)). That way variable assignments won't propagate outside the call.

effectful/handlers/llm/evaluation.py
```diff
@@ -829,3 +883,58 @@ def exec(
...
+        # via `self.write`, which `exec_code` has redirected into the buffer.
+        try:
+            exec(code, self.locals)
+        except SystemExit:
+            raise
+        except:
+            self.showtraceback()
+
+    @Tool.define
+    def exec_code(self, source: EncodableCode) -> str:
```
eb8680: As mentioned, the type of `source` should be `types.CodeType`, `EncodableCode` should be an internal implementation detail. Also let's rename `source` to `code` to match `runcode`.

effectful/handlers/llm/evaluation.py
```diff
@@ -829,3 +883,58 @@ def exec(
...
+        """Execute an `EncodableCode` and return the output it produced.
```
eb8680: The behavior of this tool differs somewhat from stateless execution tools, so this docstring should probably be much more descriptive to make sure the specification is completely clear to the LLM.

effectful/handlers/llm/evaluation.py
```diff
@@ -829,3 +883,58 @@ def exec(
...
+    output it produced; a snippet that raises has its traceback appended to that
+    output rather than propagating -- mirroring `code.InteractiveInterpreter`,
+    only `SystemExit` propagates -- so failures are surfaced as text.  There is
+    no bare-expression auto-echo, so use `print()` to surface values.
+
+    Compilation -- and therefore syntax checking -- happens earlier, when the
+    `EncodableCode` is built; this session only executes.
+    """
+
+    def __init__(self, env: Mapping[str, Any]):
```
eb8680: Sort of a nit: `ReplSession.__init__` previously overwrote `self.compiler` with a custom `codeop.CommandCompiler` subclass that invoked our `compile` and `parse` `Operation`s. I think you should reinstate that just so the behavior of `ReplSession.runsource` is fully self-consistent with `ReplSession.runcode`.

eb8680 (approved):
Looks good, thanks!

PR #691, opened 2026-06-16
Coproduct respects identity

### LINKED ISSUE #689: `{}` should be the identity for `coproduct`
These should be equivalent: `coproduct({}, i) == coproduct(i, {}) == i`.

Currently, they are not:

```
from effectful.ops.semantics import coproduct, evaluate, fwd, handler
from effectful.ops.types import Operation

f = Operation.define(int)
t = f()
i = {f: fwd}
j = {f: lambda: 42}

print(handler(coproduct(j, i))(evaluate)(t)) # 42
print(handler(coproduct(j, coproduct({}, i)))(evaluate)(t)) # int()
print(handler(coproduct(j, coproduct(i, {})))(evaluate)(t)) # 42
```

jfeser:
Closes #689

### DIFF
--- a/effectful/ops/types.py
+++ b/effectful/ops/types.py
@@ -77,6 +77,7 @@ class Operation[**Q, V]:
     __name__: str
     __default__: Callable[Q, V]
     __apply__: typing.ClassVar["Operation"]
+    fwd: "Operation"
 
     def __init__(self, default: Callable[Q, V], name: str | None = None):
         functools.update_wrapper(self, default)
@@ -497,13 +498,19 @@ class Operation[**Q, V]:
             return self
 
     def __call__(self, *args: Q.args, **kwargs: Q.kwargs) -> V:
-        from effectful.internals.runtime import get_interpretation
+        from effectful.internals.runtime import _restore_args, get_interpretation
+        from effectful.ops.semantics import fwd, handler
 
         intp = get_interpretation()
 
         self_handler = intp.get(self)
         if self_handler is not None:
-            return self_handler(*args, **kwargs)
+            # ensure that fwd is bound to the default rule. if this handler has
+            # a bound fwd, it will override this binding
+            fwd_intp = typing.cast(
+                Interpretation, {fwd: _restore_args(self.__default_rule__)}
+            )
+            return handler(fwd_intp)(self_handler)(*args, **kwargs)
         elif args and isinstance(args[0], Operation) and self is args[0].__apply__:
             # Prevent infinite recursion when calling self.apply directly
             return self.__default__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


effectful/ops/types.py
```diff
@@ -77,6 +77,7 @@ class Operation[**Q, V]:
     __name__: str
     __default__: Callable[Q, V]
     __apply__: typing.ClassVar["Operation"]
+    fwd: "Operation"
```
eb8680: Should be removed

effectful/ops/types.py
```diff
@@ -497,13 +498,29 @@ def _instance_op(instance, *args, **kwargs):
...
         intp = get_interpretation()
 
         self_handler = intp.get(self)
         if self_handler is not None:
-            return self_handler(*args, **kwargs)
+            # these cases improve performance of fwd but do not change its behavior
+            if self == _get_args:
+                return self_handler(*args, **kwargs)
+            if self == fwd:
+                return _restore_args(self_handler)(*args, **kwargs)
```
eb8680: Is this `_restore_args` always correct? Could it not restore the wrong args in some cases? I think you could also remove it with no performance cost if you use `with handler(fwd_intp)` below

effectful/ops/types.py
```diff
@@ -497,13 +498,29 @@ def _instance_op(instance, *args, **kwargs):
...
+
+            # ensure that fwd is bound to the default rule. if this handler has
+            # a bound fwd, it will override this binding
+            fwd_intp = typing.cast(
+                Interpretation, {fwd: _restore_args(self.__default_rule__)}
+            )
+            return handler(fwd_intp)(self_handler)(*args, **kwargs)
```
eb8680: Would it be cheaper to use `with handler(fwd_intp)` here instead of creating a closure?

effectful/ops/types.py
```diff
@@ -497,13 +498,29 @@ def _instance_op(instance, *args, **kwargs):
...
+        from effectful.ops.semantics import fwd, handler
 
         intp = get_interpretation()
 
         self_handler = intp.get(self)
         if self_handler is not None:
-            return self_handler(*args, **kwargs)
+            # these cases improve performance of fwd but do not change its behavior
+            if self == _get_args:
+                return self_handler(*args, **kwargs)
```
eb8680: If you use `with handler(fwd_intp)` below, this should be unnecessary

eb8680 (commented):
Looks good, a couple nits.

Also, I think we may be able to get some extra performance savings by having `_restore_args` and `_save_args` introspect their function inputs and return immediately if they have empty signatures, or if we do that introspection in `coproduct` where we still have access to the `Operation` (which is guaranteed to have a `__signature__`).

effectful/ops/types.py
jfeser: I don't see a significant difference between the forms. The cost is primarily the coproduct in `handler`.

effectful/ops/types.py
jfeser: The optimization in `_save_args/_restore_args` makes these cases unnecessary, so I've dropped them.

eb8680 (approved):

Issue #682, opened 2026-06-11
Traversal logic in `evaluate` and other core routines should be shared


eb8680:
There are currently a few different places in core library code that implement generic data structure traversals that really ought to be unified to ensure correctness and consistency of behavior:
- `ops.semantics.evaluate`
- `internals.product_n.map_structure`
- `internals.unification.nested_type`
- `ops.syntax.Scoped.analyze`
- `ops.types.Term.__str__`
- PyTree APIs in PyTorch and JAX via `handlers.torch`/`handlers.jax`

To start out, it should be relatively straightforward to refactor `evaluate` and `map_structure` to share their `singledispatch`-based traversal logic for non-`Term` data structures. Using the same traversal code in `nested_type` may be feasible as well, as should integrating with `jax.tree.map` for PyTree types.

jfeser:
Some of these uses could become calls to evaluate if we implement #715 (e.g. `Term.__str__`, `Scoped.analyze`).

PR #692, opened 2026-06-17
Implement plated einsum

jfeser:


effectful/handlers/jax/monoid.py
```diff
@@ -271,26 +362,6 @@ def _range_stop(term: Term):
...
-    """delta(i1, delta(i2, weight)) ≡ delta(i1 ++ i2, weight)"""
-
-    @implements(delta)
-    def _(self, index, weight):
-        if isinstance(weight, Term) and weight.op == delta:
-            return delta(index + weight.args[0], weight.args[1])
-        return fwd()
-
-
 class ReduceDeltaSimpleRange(ObjectInterpretation):
```
eb8680, line 391: Can we decompose `ReduceDeltaSimpleRange` into an equational rewrite step in `ops.monoid` and simpler kernel invocations in `handlers.jax.monoid`? I think this and `ReduceDependentRangeMask` are the last rules in `jax.monoid` that do any significant non-JAX-specific rewriting.

effectful/handlers/jax/monoid.py
eb8680: Is this comment obsolete with the new `mask` semantics? Either way it should probably be updated and moved or removed so it doesn't go stale.

effectful/handlers/jax/monoid.py
```diff
@@ -99,6 +104,7 @@ class ProductPlusJax(ObjectInterpretation):
     @implements(Product.plus)
     def plus(self, *args):
         if not _jax_args(args):
+            breakpoint()
```
eb8680: Remove `breakpoint`

effectful/handlers/jax/monoid.py
```diff
@@ -127,32 +133,34 @@ def plus(self, *args):
...
-    @implements(CartesianProduct.plus)
+class AndPlusJax(ObjectInterpretation):
+    @implements(And.plus)
     def plus(self, *args):
-        # Skip identity ``[()]`` args; short-circuit on zero ``[]``. Both
-        # sentinels arrive as Python lists alongside jax-array factors, so
-        # check for them explicitly before composing.
-        if not any(isinstance(a, jax.Array) for a in args):
+        if not _jax_args(args):
+            breakpoint()
```
eb8680: Remove `breakpoint`

effectful/handlers/jax/monoid.py
eb8680: Can we decompose `ReduceDependentRangeMask` into an equational rewrite step in `ops.monoid` and simpler kernel invocations in `handlers.jax.monoid`? Is this even still necessary following the move to `mask`? I think this and `ReduceDeltaSimpleRange` are the last rules in `jax.monoid` that do any significant non-JAX-specific rewriting.

tests/test_ops_syntax.py
```diff
@@ -231,6 +231,30 @@ def test_deffn_keyword_args():
     assert isinstance(result2, Term)
 
 
+def test_deffn_partial_application():
```
eb8680: Can this behavior be removed? AFAICT it's not used in the latest code, and as discussed in #703 it comes with some design issues that might be tricky to work around.

effectful/ops/monoid.py
```diff
@@ -923,41 +1315,326 @@ def reduce(self, monoid, body, streams):
...
+Termination rests on per-family progress (each "lowering" rule strictly
+consumes a resource -- a cartesian/weighted/union/array/singleton stream, a
+nested reduce/mask/delta, or a liftable factor) plus redex-shape disjointness
+that keeps the families from cycling (e.g. :class:`ReduceSplit` deliberately
+does not fire when a stream is used by *every* summand, leaving those for
+:class:`ReduceDistributeCartesianProduct`/:class:`ReduceFactorization`). There
+is no single global measure and no confluence theorem: the system is
+normalizing by construction, so a new rule that overlaps an existing redex shape
+can introduce a loop or shift which normal form is reached. Note also that
+:class:`PlusDistr` is exponential in the number of distributed sums.
```
eb8680, line 1892: Does this suggest we should consider removing `PlusDistr` or making it more conservative? I wonder what implications that would have for normalization? 

For example, without `PlusDistr`  the lhs of the following equation would no longer be rewritten to the rhs, and normalization would get stuck on the lhs, which seems bad:
```
sum(a(x) * b(y) * (c(x) + d(y)) for x in xs for y in ys) == \
  sum(b(y) * sum(a(x) * c(x) for x in xs) for y in ys) + \
  sum(a(x) * sum(b(y) * d(y) for y in ys) for x in xs)
```
On the other hand, `PlusDistr` would currently rewrite the following equation's lhs to its rhs:
```
(a + b) * (c + d) * (e + f) == \
  (a * c * e) + (a * c * f) + (a * d * e) + (a * d * f) + (b * c * e) + (b * c * f) + (b * d * e) + (b * d * f)
```
This does seem bad in isolation, but perhaps it's still worth doing if it's in the body of a `Sum.reduce` and therefore unlocks new opportunities for distributing the outer `reduce`.

effectful/ops/monoid.py
```diff
@@ -532,74 +836,147 @@ class ReduceDistributeCartesianProduct(ObjectInterpretation):
...
+            except InvalidIndexError:
+                continue
+
+            peeled_body = Union.reduce(
+                [Union.delta(drop_elem(idx, plate_index), union_body)], union_streams
+            )
+            peeled_cprod_streams = {
+                k: v for (k, v) in cprod_streams.items() if k != plate_op
+            }
+            if not peeled_cprod_streams:
```
eb8680: nit: let the generic `reduce(body, {})` rule deal with this conditional instead of inlining it, in the interest of slightly simplifying an already quite complex rule body.

effectful/ops/monoid.py
```diff
@@ -532,74 +836,147 @@ class ReduceDistributeCartesianProduct(ObjectInterpretation):
...
+            else:
+                peeled_cprod = CartesianProduct.reduce(
+                    peeled_body, peeled_cprod_streams
+                )
+
+            # include any extra inner product streams
+            inner_tail_streams = {
+                k: v for (k, v) in inner_streams.items() if k != inner_plate_op
+            }
+            if inner_tail_streams:
```
eb8680: nit: let the generic `reduce(body, {})` rule deal with this conditional instead of inlining it, in the interest of slightly simplifying an already quite complex rule body.

effectful/ops/monoid.py
```diff
@@ -532,74 +836,147 @@ class ReduceDistributeCartesianProduct(ObjectInterpretation):
...
+
+            peeled_reduce = inner_monoid.reduce(
+                monoid.reduce(inner_reduce, {stream_key: peeled_cprod}),
+                {k: v for (k, v) in inner_streams.items() if k == inner_plate_op},
             )
-            return outer_sum
+
+            # include any extra sum streams outermost
+            tail_streams = {k: v for (k, v) in streams.items() if k != stream_key}
+            if tail_streams:
```
eb8680, line 1038: nit: let the generic `reduce(body, {})` rule deal with this conditional instead of inlining it, in the interest of slightly simplifying an already quite complex rule body.

effectful/ops/monoid.py
```diff
@@ -390,16 +628,58 @@ def reduce(self, monoid, body, streams):
...
+            def group(args):
+                return args[0] if len(args) == 1 else monoid.plus(*args)
+
+            using = [arg for arg, used in zip(body.args, users) if used]
+            rest = [arg for arg, used in zip(body.args, users) if not used]
+            body2 = monoid.plus(
+                monoid.reduce(group(using), inner_streams),
+                monoid.reduce(group(rest), inner_streams),
+            )
+            return monoid.reduce(body2, outer_streams) if outer_streams else body2
```
eb8680: nit: let the generic `reduce(body, {})` rule deal with this conditional instead of inlining it, in the interest of slightly simplifying an already quite complex rule body.

effectful/ops/monoid.py
```diff
@@ -532,74 +836,147 @@ class ReduceDistributeCartesianProduct(ObjectInterpretation):
...
+                return fwd()
+
+            row_subst = {_ArrayTerm.__getitem__: _getitem}
+            try:
+                subst_inner_body = handler(row_subst)(evaluate)(inner_body)
+            except InvalidIndexError:
+                continue
+
+            peeled_body = Union.reduce(
+                [Union.delta(drop_elem(idx, plate_index), union_body)], union_streams
```
eb8680: It looks like this is the only place in `ops.monoid` where a fresh `delta` term is introduced. Is there a way to express the rule in terms of `mask` instead? It might be helpful for limiting the complexity of the normalization rules to avoid introducing `delta`s ourselves, even if we keep it in the `Monoid` interface.

eb8680 (commented):
Getting close! Seems like `mask` has clarified things.

effectful/handlers/jax/monoid.py
```diff
@@ -251,15 +250,107 @@ def reduce(self, monoid, body, streams):
...
+
+        match mask:
+            case Term(And.plus, mask_elems, {}):
+                ...
+            case _:
+                mask_elems = (mask,)
+
+        for i, elem in enumerate(mask_elems):
+            match elem:
+                case Term(_NumberTerm.__ne__, (Term(stream_op, (), {}), index)) if (
```
eb8680: Should this part of `ReduceArrayScan` be a separate rule in `ops.monoid`? The next case is array-specific, but this one isn't.

effectful/ops/monoid.py
jfeser: This conditional exists because `reduce(body, {})` doesn't have the right behavior to be a base case.

effectful/ops/monoid.py
jfeser: I don't see a way to do that.

tests/test_ops_syntax.py
jfeser: Yes.

effectful/ops/monoid.py
```diff
@@ -223,6 +263,211 @@ def _is_monoid_weighted(op: Operation) -> bool:
...
+    return isinstance(owner, Monoid) and op is owner.mask
+
+
+def _is_monoid_delta(op: Operation) -> bool:
+    """True if ``op`` is the ``delta`` operation of some :class:`Monoid`."""
+    owner = getattr(op, "__self__", None)
+    return isinstance(owner, Monoid) and op is owner.delta
+
+
+def _leaf_vars(expr: Any) -> set[Operation]:
```
eb8680: Rather than implement separate `Term` traversal logic (#682), this should post-process the result of `fvsof` to remove `Operation`s with non-empty `__signature__.parameters` - that should be faster, safer and more general.

effectful/ops/monoid.py
jfeser: What we can do instead is replace `Union.delta` with a different effectful mapping constructor.

effectful/handlers/jax/monoid.py
jfeser: Done.

effectful/handlers/jax/monoid.py
jfeser: I'd like to replace `ReduceDeltaSimpleRange` with a variant rule that eliminates equality masks, but in a followup PR.

effectful/handlers/jax/monoid.py
jfeser: We might want a rule for commuting masks through operations. The merging mentioned here is handled by mask fusion.

effectful/ops/monoid.py
jfeser: I'd like to change this behavior, but in a followup PR.

effectful/handlers/jax/monoid.py
jfeser: Done.

eb8680 (approved):

PR #703, opened 2026-06-24
Unsupplied parameters stay bound in deffn

### LINKED ISSUE #702: `deffn` should raise when passed incomplete arguments
`deffn(x() + y(), x, y)(0)` should ~evaluate to `deffn(0 + y(), y)` or~ raise an error, because this is consistent with Python's behavior for callables. It currently evaluates to `0 + y()`.

jfeser:
Closes #702

### DIFF
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -681,7 +681,14 @@ class _CallableTerm[**P, T](_BaseTerm[collections.abc.Callable[P, T]]):
                 },
             }
             with handler(subs):
-                return evaluate(body)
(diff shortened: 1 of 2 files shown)


eb8680 (commented):
I can see this being useful behavior, but it should probably be imposed consistently across the `deffn` and non-`deffn` branches. I'm also worried this will break `typeof`, since the type signature of `__call__` does not know about currying.

jfeser:
Yeah, on second thought I think I prefer to raise here. That keeps us consistent with Python callables.

Issue #690, opened 2026-06-16
REPL tool code should be typechecked when possible


eb8680:
#687 added a code execution tool for `Template`s. It would be good to typecheck the code generated for this tool when doing so is feasible, as we currently do for `Callable` result types.

Issue #577, opened 2026-02-18
Drop support for synthesis of local/runtime types


kiranandcode:
The type-checking mechanisms implemented in effectful have partial coverage for runtime-only types, but this support is partial and ad-hoc. #576 

We should drop this support and raise an exception, and add special case handling for notebooks explicitly.

eb8680:
I think this was effectively resolved by #706

Issue #576, opened 2026-02-17
Decoding runtime-only Pydantic models in Callables fails


kiranandcode:
If a pydantic model is only available at runtime and can not be imported from a file, then it is not possible to create a Callable that uses this model:

```python
from collections.abc import Callable

import pydantic

from effectful.handlers.llm.encoding import Encodable, SynthesizedFunction
from effectful.handlers.llm.evaluation import UnsafeEvalProvider
from effectful.ops.semantics import handler


# Dynamically create a pydantic model (simulates runtime-generated models)
def _make_model():
    class MyModel(pydantic.BaseModel):
        x: int
        y: str
    return MyModel

MyModel = _make_model()

typ = Callable[[MyModel], MyModel]
source = SynthesizedFunction(
    module_code="def f(a: MyModel) -> MyModel:\n    return a"
)
ctx = {"MyModel": MyModel}
enc = Encodable.define(typ, ctx)

with handler(UnsafeEvalProvider()):
    f = enc.decode(source)  # fails: mypy [override] error from generated stubs
    assert f(MyModel(x=1, y="hi")) == MyModel(x=1, y="hi")

```

Failing with issue:
```
Traceback (most recent call last):
  File "<string>", line 17, in __PYTHON_EL_eval
  File "/Users/kirancodes/Documents/code/effectful/tmp.py", line 27, in <module>
    f = enc.decode(source)  # fails: mypy [override] error from generated stubs
        ^^^^^^^^^^^^^^^^^^
  File "/Users/kirancodes/Documents/code/effectful/effectful/handlers/llm/encoding.py", line 451, in decode
    evaluation.type_check(
  File "/Users/kirancodes/Documents/code/effectful/effectful/ops/types.py", line 494, in __call__
    return self_handler(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kirancodes/Documents/code/effectful/effectful/internals/runtime.py", line 70, in bound_body
    return body(*a, **k)
           ^^^^^^^^^^^^^
  File "/Users/kirancodes/Documents/code/effectful/effectful/internals/runtime.py", line 56, in _cont_wrapper
    return fn(*a, **k)
           ^^^^^^^^^^^
  File "/Users/kirancodes/Documents/code/effectful/effectful/handlers/llm/evaluation.py", line 689, in type_check
    mypy_type_check(module, ctx, expected_params, expected_return)
[...]

Issue #542, opened 2026-02-06
Synthesized Callable Name shouldn't matter


datvo06:
After #536, we have mypy check that check redefinition. This raise trouble in some case:
```python
import inspect
from collections.abc import Callable

from effectful.handlers.llm.evaluation import UnsafeEvalProvider
from effectful.handlers.llm import Template
from effectful.ops.semantics import NotHandled, handler
from effectful.handlers.llm.completions import LiteLLMProvider, RetryLLMHandler


provider = LiteLLMProvider()


@Template.define
def count_char(char: str) -> Callable[[str], int]:
    """Write a function which takes a string and counts the occurrances of '{char}'. Do not use any tools.
    Remember to name the function as count_{char}.
    """
    raise NotHandled


# Use UnsafeEvalProvider for simple examples; RestrictedEvalProvider may need extra globals.
with handler(provider), handler(UnsafeEvalProvider()), handler(RetryLLMHandler(num_retries=3)):
    count_a = count_char("a")
    assert callable(count_a)
    assert count_a("banana") == 3
    assert count_a("cherry") == 0
    # Print the source code of the generated function
    print(inspect.getsource(count_a))

    count_a = count_char("a")
    assert callable(count_a)
    assert count_a("banana") == 3
    assert count_a("cherry") == 0
    # Print the source code of the generated function
    print(inspect.getsource(count_a))
```

PR #706, opened 2026-06-30
Replace runtime-type quotation with splice-into-source typechecking

### LINKED ISSUE #696: Round-tripping through quotation in `handlers.llm` is brittle and incomplete
There are a few ostensibly general behaviors in `handlers.llm` that currently rely on quoting runtime types and values (generating code or strings that can be decoded losslessly to their original values). 

Most notably, our code generation functionality in `handlers.llm` currently includes a type-checking stage that is context-aware, which is great in theory, but the implementation depends on being able to generate an `ast.AST` for a variable annotation statement (`x: int`, etc) from a runtime type. The type quotation mechanism in `handlers.llm.evaluation` is buggy and incomplete (basic things like `Enum` fail, sometimes silently), and is inherently doomed to remain so to some extent given how broad the language is.

It follows that rather than just attempting to fix these failures piecemeal, we should perhaps consider alternative approaches that don't require quotation in the first place. For example, in type-checking generated `Callable`s we could splice the generated code into the source code or AST of the module at the location where a `Template`'s body would be defined, and type-check the entire modified file. Similar observations apply for the related issues #690 #577 and maybe even #489 to some extent

datvo06:
Closes #696. Addresses #577, #576.

Currently, for synthesis, we are reconstructing a `Template`'s lexical context as synthetic type stubs. This is brittle and likely unsound.

This PR splices the generated code into the Template module, and type checks the whole file. This come with a few changes

1. `mypy_type_check` now parameterized with `generated` (code) and `anchor` (the Template). The anchor is used to recover its module source, 

Note, in practice there is a few more tweaks: 
-  Mypy runs on the whole module, but only diagnostics inside the spliced span are raised, so an unrelated pre-existing error elsewhere in the module can't fail synthesis. 
- Non-nestable constructs (`from x import *`, `__future__`) are caught by an explicit AST pre-scan, since mypy silently accepts a nested star-import; a fatal mypy error (exit ≥ 2) is surfaced rather than silently passed; unrecoverable source (no file, no `linecache`) is skipped with a logged reason.

2. **Name collisions (#542)** are now handled structurally: the generated def is a function-body local that *shadows* a colliding context name rather than redefining it at module level, so the AST-rename pass is gone. The whole quotation layer is deleted — `type_to_ast`, `signature_to_ast`, `collect_variable_declarations`, `collect_runtime_type_stubs`, `collect_imports`, and `_RenameTransformer`/`_generate_unique_name` (net +510 / −1991).

3. **Behavior change.** Runtime-only / local types, and names with no lexical presence at the splice point, now raise instead of being (partially, fakily) stubbed. The check is sound on annotated generated code; unannotated bodies are checked at their signature only (`--check-untyped-defs` is deliberately off — it rejects runnable heterogeneous-container code).

**Tests.** 
1. Deleted the quotation-internals unit tests; kept the security/REPL tests
2. Added behavior tests asserting raise-or-not (not mypy text): #542 collision shadowing, the gate (unrelated module error doesn't block a correct synthesis), closures (enclosing locals in scope), method/static/class templates, `Enum` + dataclass context (#576), injected-name, non-nestable, `linecache`-backed recovery (the #690 path), annotation-collision, and end-to-end decode through `Encodable[Callable]`. `mypy effectful/handlers/llm/` clean, `ruff` clean, full LLM suite green.

### DIFF
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -30,7 +30,7 @@ from effectful.handlers.llm.encoding import (
     Encodable,
     to_content_blocks,
 )
-from effectful.handlers.llm.evaluation import ReplSession
+from effectful.handlers.llm.evaluation import ReplSession, type_check_anchor
 from effectful.handlers.llm.template import (
     Agent,
     Template,
@@ -430,15 +430,18 @@ def call_assistant[T](
     encoding: pydantic.TypeAdapter[DecodedToolCall] = pydantic.TypeAdapter(
         Encodable[DecodedToolCall]
     )
-    for raw_tool_call in message.get("tool_calls") or []:
-        try:
-            tool_calls += [encoding.validate_python(raw_tool_call, context=tools)]
-        except Exception as e:
-            raise ToolCallDecodingError(
-                raw_tool_call=raw_tool_call,
-                original_error=e,
-                raw_message=raw_message,
-            ) from e
+    # Tool-argument Callables have no source anchor (out of scope for the splice
+    # type check), so suppress the anchor while decoding them.
+    with handler({type_check_anchor: lambda: None}):
+        for raw_tool_call in message.get("tool_calls") or []:
+            try:
+                tool_calls += [encoding.validate_python(raw_tool_call, context=tools)]
+            except Exception as e:
+                raise ToolCallDecodingError(
+                    raw_tool_call=raw_tool_call,
+                    original_error=e,
+                    raw_message=raw_message,
+                ) from e
 
     result = None
     if not tool_calls:
@@ -672,7 +675,15 @@ class LiteLLMProvider(ObjectInterpretation):
         )  # type: ignore
         history_copy = history.copy()
 
-        with handler({_get_history: lambda: history_copy}):
+        # Bind the type-check anchor to this Template's underlying function for the
+        # duration of the call. Handler scope (not data context) means nested or
+        # recursive synthesis sees the right anchor -- innermost wins.
+        with handler(
+            {
+                _get_history: lambda: history_copy,
+                type_check_anchor: lambda: template.__default__,
+            }
+        ):
             if (
                 not _get_history()
                 or next(iter(_get_history().values()))["role"] != "system"
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -430,8 +430,11 @@ def call_assistant[T](
     encoding: pydantic.TypeAdapter[DecodedToolCall] = pydantic.TypeAdapter(
         Encodable[DecodedToolCall]
     )
-    # Tool-argument Callables have no source anchor (out of scope for the splice
-    # type check), so suppress the anchor while decoding them.
+    # A synthesized tool-argument Callable's contract is the tool parameter's
+    # type, not the enclosing Template's return type -- so the Template anchor
+    # (bound by __apply__) is the wrong splice target for it, and it has no def of
+    # its own to splice into. Suppress the anchor to skip the check while decoding
+    # tool arguments, rather than check them against the wrong type.
     with handler({type_check_anchor: lambda: None}):
         for raw_tool_call in message.get("tool_calls") or []:
             try:
--- a/effectful/handlers/llm/completions.py
+++ b/effectful/handlers/llm/completions.py
@@ -26,11 +26,12 @@ from litellm import (
 )
 
 from effectful.handlers.llm.encoding import (
+    TYPE_CHECK_ANCHOR_KEY,
     DecodedToolCall,
     Encodable,
     to_content_blocks,
 )
-from effectful.handlers.llm.evaluation import ReplSession, type_check_anchor
+from effectful.handlers.llm.evaluation import ReplSession
 from effectful.handlers.llm.template import (
     Agent,
     Template,
@@ -392,6 +393,7 @@ def call_assistant[T](
         ResultDecodingError: If the result cannot be decoded. The error
             includes the raw assistant message for retry handling.
     """
+    anchor = kwargs.pop("anchor", None)  # ride in kwargs; pop before the LLM call
     tools = dict(collect_tools(env))
     tool_specs = {
         k: typing.cast(
@@ -430,21 +432,19 @@ def call_assistant[T](
     encoding: pydantic.TypeAdapter[DecodedToolCall] = pydantic.TypeAdapter(
         Encodable[DecodedToolCall]
     )
-    # A synthesized tool-argument Callable's contract is the tool parameter's
-    # type, not the enclosing Template's return type -- so the Template anchor
-    # (bound by __apply__) is the wrong splice target for it, and it has no def of
-    # its own to splice into. Suppress the anchor to skip the check while decoding
-    # tool arguments, rather than check them against the wrong type.
-    with handler({type_check_anchor: lambda: None}):
-        for raw_tool_call in message.get("tool_calls") or []:
-            try:
-                tool_calls += [encoding.validate_python(raw_tool_call, context=tools)]
-            except Exception as e:
-                raise ToolCallDecodingError(
-                    raw_tool_call=raw_tool_call,
-                    original_error=e,
-                    raw_message=raw_message,
-                ) from e
+    # Tool arguments decode with `context=tools` (no anchor key), so a synthesized
+    # tool-argument Callable gets no type-check anchor -- correctly: its contract
+    # is the tool parameter's type, not the enclosing Template's return type, so
+    # the Template anchor would be the wrong splice target.
+    for raw_tool_call in message.get("tool_calls") or []:
+        try:
+            tool_calls += [encoding.validate_python(raw_tool_call, context=tools)]
+        except Exception as e:
+            raise ToolCallDecodingError(
+                raw_tool_call=raw_tool_call,
+                original_error=e,
+                raw_message=raw_message,
+            ) from e
 
     result = None
     if not tool_calls:
@@ -457,8 +457,12 @@ def call_assistant[T](
             result = typing.cast(T, serialized_result)
         else:
             try:
+                # Add the type-check anchor to the decode context only (not `env`,
+                # which is exposed as tools), so a synthesized result is checked
+                # against the Template's source.
                 result = response_format.model_validate(
-                    json.loads(serialized_result), context=env
+                    json.loads(serialized_result),
+                    context={**env, TYPE_CHECK_ANCHOR_KEY: anchor},
                 ).value
             except Exception as e:
                 raise ResultDecodingError(e, raw_message=raw_message) from e
@@ -678,15 +682,7 @@ class LiteLLMProvider(ObjectInterpretation):
         )  # type: ignore
         history_copy = history.copy()
 
-        # Bind the type-check anchor to this Template's underlying function for the
-        # duration of the call. Handler scope (not data context) means nested or
-        # recursive synthesis sees the right anchor -- innermost wins.
-        with handler(
-            {
-                _get_history: lambda: history_copy,
-                type_check_anchor: lambda: template.__default__,
-            }
-        ):
+        with handler({_get_history: lambda: history_copy}):
             if (
                 not _get_history()
                 or next(iter(_get_history().values()))["role"] != "system"
@@ -700,7 +696,10 @@ class LiteLLMProvider(ObjectInterpretation):
             result: T | None = None
             while message["role"] != "assistant" or tool_calls:
                 message, tool_calls, result = call_assistant(
-                    env, template.__signature__.return_annotation, **self.config
+                    env,
+                    template.__signature__.return_annotation,
+                    anchor=template.__default__,
+                    **self.config,
                 )
                 for tool_call in tool_calls:
                     message = call_tool(tool_call)
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -39,6 +39,14 @@ from effectful.ops.types import Operation, Term
 
 type ToolCallID = str
 
+# Reserved key under which the type-check anchor (the enclosing Template's
+# underlying function) rides in the Pydantic decoding context, alongside the
+# lexical environment. `decode` reads it to type-check a synthesized function
+# against the Template's source; absent (tool-argument decoding) means skip.
+# Deliberately not a valid identifier so `LexicalReaders` skips it (no tool leak)
+# and it can never collide with a lexical name.
+TYPE_CHECK_ANCHOR_KEY = "<type_check_anchor>"
+
 CONTENT_BLOCK_TYPES: frozenset[str] = frozenset(
     literal
     for member in typing.get_args(OpenAIMessageContentListBlock)
@@ -552,19 +560,19 @@ def _pydantic_callable(callable_type: Any) -> Any:
 
         _validate_signature_ast(last_stmt, expected_params)
 
-        # `type_check_anchor` is the Template's underlying function (bound per call
-        # by Template.__apply__); None for tool-argument decoding, whose synthesized
-        # Callables are contracted by the tool param's type, not the Template's
-        # return type, so the Template anchor doesn't apply. When present, the code
-        # is spliced into the Template body, so first reject constructs illegal once
-        # nested (star / `__future__` imports), then type-check.
-        anchor = evaluation.type_check_anchor()
+        # The anchor (Template's underlying function) rides in the decoding context
+        # under TYPE_CHECK_ANCHOR_KEY; absent for tool-argument decoding, whose
+        # synthesized Callables are contracted by the tool param's type, not the
+        # Template's return type, so the Template anchor doesn't apply. When
+        # present, the code is spliced into the Template body, so first reject
+        # constructs illegal once nested (star / `__future__` imports), then check.
+        anchor = ctx.get(TYPE_CHECK_ANCHOR_KEY)
         if anchor is not None:
             evaluation.scan_non_nestable(module)
             evaluation.type_check(module, anchor)
 
         g: MutableMapping[str, Any] = {}
-        g.update(ctx)
+        g.update({k: v for k, v in ctx.items() if k != TYPE_CHECK_ANCHOR_KEY})
         bytecode: types.CodeType = evaluation.compile(module, filename)
         evaluation.exec(bytecode, g)
 
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -569,10 +569,18 @@ def _pydantic_callable(callable_type: Any) -> Any:
         anchor = ctx.get(TYPE_CHECK_ANCHOR_KEY)
         if anchor is not None:
             evaluation.scan_non_nestable(module)
-            evaluation.type_check(module, anchor)
+            spliced = evaluation.splice_into_source(module, anchor)
+            if spliced is not None:
+                evaluation.type_check(*spliced)
 
         g: MutableMapping[str, Any] = {}
-        g.update({k: v for k, v in ctx.items() if k != TYPE_CHECK_ANCHOR_KEY})
+        g.update(
+            {
+                k: v
+                for k, v in ctx.items()
+                if k.isidentifier() and k != TYPE_CHECK_ANCHOR_KEY
+            }
+        )
         bytecode: types.CodeType = evaluation.compile(module, filename)
         evaluation.exec(bytecode, g)
 
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -551,7 +551,13 @@ def _pydantic_callable(callable_type: Any) -> Any:
             )
 
         _validate_signature_ast(last_stmt, expected_params)
-        evaluation.type_check(module, ctx, expected_params, expected_return)
+
+        # `type_check_anchor` is the Template's underlying function, bound per call
+        # by Template.__apply__; it is None for tool-argument decoding, which
+        # therefore skips the source-anchored type check.
+        anchor = evaluation.type_check_anchor()
+        if anchor is not None:
+            evaluation.type_check(module, anchor)
 
         g: MutableMapping[str, Any] = {}
         g.update(ctx)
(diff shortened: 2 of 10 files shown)


datvo06:
My intuition is mypy failed with stress test. Checking..

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+    if template_def is None:
+        logger.warning(
+            "skipping type check: cannot locate %r in its module source",
+            getattr(fn, "__qualname__", fn),
+        )
+        return None
+
+    # Splice in place: replace the body with the generated body, binding the
+    # target against the (source) return annotation via `return`. Strip the
+    # app-level decorators (`@Template.define`, …) but KEEP `@staticmethod`/
```
eb8680: Why do we need to strip the decorators at all, if the code being spliced is entirely confined to the function body? The less we do to the surrounding code, the more robust this splicing procedure will be.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+
+    fn = _unwrap(anchor)
+    module_source = _recover_module_source(fn)
+    if module_source is None:
+        logger.warning("skipping type check: cannot recover source for %r", fn)
+        return None
+    try:
+        module_ast = ast.parse(module_source)
+    except SyntaxError:
+        logger.warning("skipping type check: cannot parse module source for %r", fn)
```
eb8680: Why would this fail, and why would we want that failure to be silent? Shouldn't we have confirmed by this point in the computation that the code is syntactically valid?

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-    if status != 0:
-        report = (stdout or "") + (stderr or "")
-        raise TypeError(f"mypy type check failed:\n{report}\n{source}")
+    checked_source = ast.unparse(ast.fix_missing_locations(module_ast))
+
+    # Locate the spliced def's line span in the *checked* source's coordinates
+    # (unparse reassigns line numbers), so diagnostics can be scoped to it.
+    checked_ast = ast.parse(checked_source)
+    spliced = _find_def_by_qualname(checked_ast, fn.__qualname__)
+    if spliced is None:  # pragma: no cover - the splice is always re-locatable
```
eb8680: What does this comment mean? When would `spliced` be `None`?

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-        ]
-    )
-    if status != 0:
-        report = (stdout or "") + (stderr or "")
-        raise TypeError(f"mypy type check failed:\n{report}\n{source}")
+    checked_source = ast.unparse(ast.fix_missing_locations(module_ast))
+
+    # Locate the spliced def's line span in the *checked* source's coordinates
+    # (unparse reassigns line numbers), so diagnostics can be scoped to it.
+    checked_ast = ast.parse(checked_source)
```
eb8680: Why are we `unparse`ing and immediately `parse`ing again here? Does this lose information? Is this doing something on top of what `ast.fix_missing_locations` is doing to `module_ast`?

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+    try:
+        module_ast = ast.parse(module_source)
+    except SyntaxError:
+        logger.warning("skipping type check: cannot parse module source for %r", fn)
+        return None
+
+    template_def = _find_def_by_qualname(module_ast, fn.__qualname__)
+    if template_def is None:
+        logger.warning(
+            "skipping type check: cannot locate %r in its module source",
```
eb8680: If we are able to recover the module's source, shouldn't `fn` have definition site line numbers associated with it that we can use to identify it directly, instead of this potentially failure-prone name matching heuristic `_find_def_by_qualname`? Under what conditions would we expect to recover `module_ast` but not be able to locate `fn`?

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-    imports = collect_imports(ctx)
-    # Ensure annotations in the postlude can be resolved (e.g. collections.abc.Callable, typing)
-    baseline_imports: list[ast.stmt] = [
-        ast.Import(names=[ast.alias(name="collections", asname=None)]),
-        ast.Import(names=[ast.alias(name="collections.abc", asname=None)]),
-        ast.Import(names=[ast.alias(name="typing", asname=None)]),
-        ast.Import(names=[ast.alias(name="types", asname=None)]),
+    target_name = last.name
+
+    _scan_non_nestable(generated)
```
eb8680: This check should probably happen prior to `type_check`

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+                found = descend(node.body, idx + 1)
+                if found is not None:
+                    return found
+        return None
+
+    return descend(module.body, 0)
+
+
+# mypy diagnostics under --no-pretty: ``file:line: error: ...`` or ``file:line:col: ...``.
+_MYPY_LINE_RE = re.compile(r"^.+?:(\d+):(?:\d+:)?\s+(?:error|note):")
```
eb8680: Where does this regex come from? Is it documented somewhere in mypy? How can we be sure it's not filtering too aggressively?

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-        ast.Import(names=[ast.alias(name="typing", asname=None)]),
-        ast.Import(names=[ast.alias(name="types", asname=None)]),
+    target_name = last.name
+
+    _scan_non_nestable(generated)
+
+    fn = _unwrap(anchor)
+    module_source = _recover_module_source(fn)
+    if module_source is None:
+        logger.warning("skipping type check: cannot recover source for %r", fn)
```
eb8680, line 219: This should probably also happen prior to `type_check`

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+    report = (stdout or "") + (stderr or "")
+    region = _region_diagnostics(report, lo, hi)
+    if any(" error:" in line for line in region):
+        raise TypeError(
+            "mypy type check failed:\n" + "\n".join(region) + "\n\n" + checked_source
+        )
+    if status >= 2:
+        # Exit code >= 2 means mypy itself failed (fatal/usage/internal error),
+        # not that it found type errors -- surface it rather than silently pass.
+        raise TypeError(f"mypy could not check the spliced module:\n{report}")
```
eb8680: This should be a `RuntimeError` or an assertion, not a `TypeError`, and the mypy status check should probably happen before you potentially raise a type-checking `TypeError` above. We should only report type errors if mypy runs to completion.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-    "_frozen_importlib",
-)
+def _unwrap(fn: Any) -> Any:
+    """Unwrap a ``staticmethod``/``classmethod`` to its underlying function."""
+    return fn.__func__ if isinstance(fn, staticmethod | classmethod) else fn
 
 
-def collect_imports(ctx: Mapping[str, Any]) -> list[ast.stmt]:
-    """Collect module imports and symbol imports from context.
+def _recover_module_source(fn: Any) -> str | None:
```
eb8680: Is this function really necessary, or can it be removed in favor of `inspect.getsource(inspect.getmodule(fn))`? The less of this munging we do ourselves, the more robust this procedure will be.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-    "__annotations__",
-    "__doc__",
-    "__file__",
-    "__loader__",
-    "__name__",
-    "__package__",
-    "__spec__",
-    "_frozen_importlib",
-)
+def _unwrap(fn: Any) -> Any:
```
eb8680: Is this really necessary, or can it be removed if we outsource more of the introspection of `fn` to `inspect`? The fewer of these heuristics we have the more robust our type-checking step will be.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
-    not Python keywords, and not in the given set of existing names.
+            with open(filename, encoding="utf-8") as f:
+                return f.read()
+        except OSError:
+            return None
+    lines = linecache.getlines(filename)
+    return "".join(lines) if lines else None
+
+
+def _find_def_by_qualname(
```
eb8680: As mentioned in another comment, it would be better to remove this heuristic in favor of exact lookup using line numbers that guarantee correctness in all cases.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +112,224 @@ def exec(
...
+    """Diagnostic lines whose reported line number falls within ``[lo, hi]``."""
+    region: list[str] = []
+    for line in report.splitlines():
+        m = _MYPY_LINE_RE.match(line)
+        if m and lo <= int(m.group(1)) <= hi:
+            region.append(line)
+    return region
+
+
+def mypy_type_check(generated: ast.Module, anchor: Any) -> None:
```
eb8680: I think we probably want to split this into two functions, one that handles splicing and constructing the modified source code and one that just applies mypy to whatever source it's given, whether it came from the splicing function or somewhere else.

effectful/handlers/llm/completions.py
```diff
@@ -430,15 +430,18 @@ def call_assistant[T](
...
-    for raw_tool_call in message.get("tool_calls") or []:
-        try:
-            tool_calls += [encoding.validate_python(raw_tool_call, context=tools)]
-        except Exception as e:
-            raise ToolCallDecodingError(
-                raw_tool_call=raw_tool_call,
-                original_error=e,
-                raw_message=raw_message,
-            ) from e
+    # Tool-argument Callables have no source anchor (out of scope for the splice
```
eb8680: What does this comment mean? When would this apply? Surely if we are synthesizing code to pass to a `Tool` we have access to its source, and we know it's happening lexically in the body of the `Template` definition?

effectful/handlers/llm/completions.py
```diff
@@ -672,7 +675,15 @@ def _call[**P, T](
...
         history_copy = history.copy()
 
-        with handler({_get_history: lambda: history_copy}):
+        # Bind the type-check anchor to this Template's underlying function for the
+        # duration of the call. Handler scope (not data context) means nested or
+        # recursive synthesis sees the right anchor -- innermost wins.
+        with handler(
+            {
+                _get_history: lambda: history_copy,
+                type_check_anchor: lambda: template.__default__,
```
eb8680: Rather than managing this synthesis context variable with an operation and handler that are hard to disentangle from `Template.__apply__`, it would probably be better to route it through Pydantic's context argument during decoding, as we already do with the lexical environment.

eb8680 (commented):
Thanks, this seems like a big improvement. There are several heuristics and silent failures that I worry will undermine its correctness, though.

effectful/handlers/llm/evaluation.py
datvo06: Earlier `Claude` convinced me with some example that if we keep the annotator, then mypy would fail. But checking again carefully, it's not the case.
I'm removing the strip decorator step.

effectful/handlers/llm/evaluation.py
datvo06: It would fail if we parsed the file while the writing was not done. I think I encountered it during one test, but I can't reproduce that since then. Will remove this.

effectful/handlers/llm/evaluation.py
datvo06: It can't be None if `_find_def_by_qualname` does not return None. I'm removing this check

effectful/handlers/llm/evaluation.py
datvo06: It needed to do it to remove the comments, otherwise we would have to do comments re-indenting.

effectful/handlers/llm/evaluation.py
datvo06: Yes, I now use `template_def = _find_def_at_lineno(module_ast, fn.__code__.co_firstlineno)`, it's the first line comment, so it's both the comment and the function body, we wouldnt have to worry about aligning comment offset.

effectful/handlers/llm/evaluation.py
datvo06: Updated!

effectful/handlers/llm/evaluation.py
datvo06: This is the example: 
```python
  def count_a(s):
      return s.count('a')
```
but the template body is indented (inside def the_template(...):, maybe inside a class too). To splice it in, you must add leading spaces to every generated line. That's fine for code, but it corrupts multi-line string literals, because the leading spaces land inside the string:
```python
  def gen():
      """line one
  line two"""          # the string is "line one\nline two" indent by 4 to place it in the body →
      def gen():
          """line one
      line two"""       # the string is now "line one\n    line two"  ← changed!
```
line two gained 4 spaces inside the docstring. The string's value changed. And LLM-generated functions very often have docstrings, so this isn't a corner case.

effectful/handlers/llm/evaluation.py
datvo06: I moved it before `type_check`

effectful/handlers/llm/evaluation.py
datvo06: It's a heuristic that is neither sound nor complete. I think from mypy 1.14 it allows output in Json so we wouldn't have to worry about parsing error, implementing it now.

datvo06: It works. Now using `--output-json` so we don't have to parse error.

effectful/handlers/llm/evaluation.py
datvo06: Updated!

effectful/handlers/llm/evaluation.py
datvo06: Updated! `inspect.getsource(inspect.getmodule(fn))` doesn't work for `REPL` or notebook case, so I used `inspect.getsourcefile(fn)` followed by reading lines from this file.

effectful/handlers/llm/evaluation.py
datvo06: I removed it in favor of `inspect.unwrap`.

effectful/handlers/llm/evaluation.py
datvo06: I split the `splice_into_source` and `_mypy_check_region(source: str, lo: int, hi: int)`.

effectful/handlers/llm/completions.py
datvo06: Sorry, I was mistaken between decoding a Tool (not possible) and encoding it. Removed the comment.

effectful/handlers/llm/evaluation.py
eb8680: OK, when could `template_def` ever be `None` now? Can we remove this `if` block entirely, so that `_splice_into_source` can fail loudly if we haven't checked this condition already?

datvo06: It happens when we have multiple edits and the file pointer is stale. For example, we import a template from `foo.py`, edit `foo.py`, then trigger synthesis without reimport, then `fn` is stale, the recovered source is new, and the def has moved. Let me make it fail loudly for now.

effectful/handlers/llm/completions.py
datvo06: Updated. 
There is still a problem with lexical environment that if we pass it into the environment, then it might pickup itself as a tool again, so I now route `anchor` through `call_assistant` and then to decoding context.

effectful/handlers/llm/encoding.py
```diff
@@ -551,10 +559,20 @@ def _validate(value: Any, info: pydantic.ValidationInfo) -> Callable:
...
+        # present, the code is spliced into the Template body, so first reject
+        # constructs illegal once nested (star / `__future__` imports), then check.
+        anchor = ctx.get(TYPE_CHECK_ANCHOR_KEY)
+        if anchor is not None:
+            evaluation.scan_non_nestable(module)
+            evaluation.type_check(module, anchor)
 
         g: MutableMapping[str, Any] = {}
-        g.update(ctx)
+        g.update({k: v for k, v in ctx.items() if k != TYPE_CHECK_ANCHOR_KEY})
```
eb8680: nit: use the future-proof `if k.isidentifier()`

effectful/handlers/llm/evaluation.py
```diff
@@ -51,23 +48,19 @@ def parse(source: str, filename: str) -> ast.Module:
 
 
 @defop
-def type_check(
-    module: ast.Module,
-    ctx: typing.Mapping[str, Any],
-    expected_params: list[type] | None,
-    expected_return: type,
-) -> None:
+def type_check(generated: ast.Module, anchor: Any) -> None:
```
eb8680: This should just take the spliced source, or the spliced AST, and perhaps optional line numbers delimiting a region. Splicing can happen separately.

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +98,224 @@ def exec(
...
+    # Exit status >= 2 means mypy itself failed (fatal/usage/internal/syntax) -- a
+    # tool failure, not a type error -- and it emits text rather than JSON, so
+    # raise `RuntimeError` rather than parse or silently pass.
+    if status >= 2:
+        raise RuntimeError(
+            f"mypy could not check the source:\n{(stdout or '') + (stderr or '')}"
+        )
+    errors = _region_errors(stdout or "", lo, hi)
+    if errors:
+        report = "\n".join(
```
eb8680: nit: just use `json.dumps(e)` to serialize each error verbatim

effectful/handlers/llm/evaluation.py
```diff
@@ -105,584 +98,224 @@ def exec(
...
+            f"{e['line']}: {e['message']}  [{e['code']}]" for e in errors
+        )
+        raise TypeError("mypy type check failed:\n" + report + "\n\n" + source)
```
eb8680: I don't think you need to include the source, which could be quite large - the model should already have the generated source in context.

effectful/handlers/llm/completions.py
```diff
@@ -686,7 +696,10 @@ def _call[**P, T](
             result: T | None = None
             while message["role"] != "assistant" or tool_calls:
                 message, tool_calls, result = call_assistant(
-                    env, template.__signature__.return_annotation, **self.config
+                    env,
+                    template.__signature__.return_annotation,
+                    anchor=template.__default__,
```
eb8680, line 701: I'm not sure `template.__default__` will work as `anchor` for bound-method `Template`s. You should make sure this case is covered in a test, which you'll probably then need to fix.

effectful/handlers/llm/encoding.py
```diff
@@ -551,10 +559,20 @@ def _validate(value: Any, info: pydantic.ValidationInfo) -> Callable:
...
+        # The anchor (Template's underlying function) rides in the decoding context
+        # under TYPE_CHECK_ANCHOR_KEY; absent for tool-argument decoding, whose
+        # synthesized Callables are contracted by the tool param's type, not the
+        # Template's return type, so the Template anchor doesn't apply. When
+        # present, the code is spliced into the Template body, so first reject
+        # constructs illegal once nested (star / `__future__` imports), then check.
+        anchor = ctx.get(TYPE_CHECK_ANCHOR_KEY)
+        if anchor is not None:
+            evaluation.scan_non_nestable(module)
+            evaluation.type_check(module, anchor)
```
eb8680: Splicing can happen just before `type_check` is called here, rather than inside `type_check`.

eb8680 (commented):
Nearly there, just a few small things

effectful/handlers/llm/encoding.py
datvo06: I added `k.isidentifier()` check along with the avoiding anchor collision check.

effectful/handlers/llm/evaluation.py
datvo06: Updated! I also removed `mypy_type_check` and now use `mypy_typecheck_region` instead.

effectful/handlers/llm/evaluation.py
datvo06: Updated! I also removed the source in the report.

effectful/handlers/llm/evaluation.py
datvo06: Updated!

effectful/handlers/llm/completions.py
datvo06: I added an integration test, and it works ok after multiple retries (due to LLM flakiness), also added the replay record.

effectful/handlers/llm/encoding.py
datvo06: Updated. Now `splice` is called before `type_check`

effectful/handlers/llm/completions.py
datvo06: The relay buffer gives me this:
```
┌───────┬────────────────────────────────────────────────────┬────────┐
│ Round │                   Generated code                   │ Valid? │
├───────┼────────────────────────────────────────────────────┼────────┤
│ 0     │ def add(a, b) -> int:\n    return a + b<br>\n\nadd │ ✗      │
├───────┼────────────────────────────────────────────────────┼────────┤
│ 1     │ def add(...): ...<br>\n\n# comment\n\nadd          │ ✗      │
├───────┼────────────────────────────────────────────────────┼────────┤
│ 2     │ def add(...): ...<br>\n\nadd                       │ ✗      │
├───────┼────────────────────────────────────────────────────┼────────┤
│ 3     │ def add(a: int, b: int) -> int:\n    return a + b  │ ✓      │
└───────┴────────────────────────────────────────────────────┴────────┘
```

The function body is correct every time — that's the part that "looks correct." But rounds 0–2 append a trailing bare add (the function name on its own line, as a "here's the result" hint). That makes the module's last statement a bare Expr, and decode requires it to be a FunctionDef:
```
Value error, decode() requires the last statement to be a function definition, got Expr
```

I think this is more of the mypy thing but not sure we should try adding fix in this PR cause
I'm not sure how sound/complete the solution could be, and it still works after multiple retries.

eb8680 (approved):

PR #708, opened 2026-07-07
Add tail call fwd interface

### LINKED ISSUE #704: Reduce stack use in deeply nested handlers
Our current approach to deferring from one handler to another is to call `fwd`, which is dynamically bound to the next relevant handler. This puts multiple frames on the stack and for deeply nested handlers (as in weighted) we quickly blow the stack on even moderately sized terms. 

However, the common calling pattern for `fwd` is that the call is in tail position, so in principle there's no need to use any additional stack frames in these cases. Python doesn't offer tail-call optimization, so we would need our own solution. 

The simplest solution that I can see is to have another pattern for forwarding where the handler would return a marker value instead of calling `fwd`. This would have the advantage of forcing use only in tail position, but makes the forwarding interface more complex.

jfeser:
Closes #704

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -40,7 +40,7 @@ def fwd(*args, **kwargs) -> Any:
 @dataclass
 class Fwd:
     next: Callable
-    args: tuple | None = ()
+    args: tuple = ()
     kwargs: collections.abc.Mapping = field(default_factory=dict)
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


cooijmanstim:
Would it make sense to make `fwd` return a `Fwd` object, so existing handlers get TCO?

jfeser:
No. `fwd` must return a real value, not a marker, so handlers that compute with the return value continue to work.

PR #711, opened 2026-07-15
Type-checking REPL

### LINKED ISSUE #690: REPL tool code should be typechecked when possible

datvo06:
Closes #690. The exec_code REPL tool (#687) runs LLM-generated code unchecked.
This PR type-checks it, reusing the type_check op from #706.

A REPL is a module built one snippet at a time, so I check each snippet against the accumulated prior snippets and write any in-region type error to stderr. 
If we run into errors, the snippet still runs (a type error isn't a runtime error, and the session's contract is that errors are text the model reads).

Several design choices I made:
1.  One challenge I got to solve is when you re-define a variable; this is a valid REPL use, but mypy would raise `no-redef`. 

A simple solution would be disabling `no-redef`, but then mypy would raise in case of something like this:
```python
def f() -> int:
    ...

# later
def f(x, y) -> int: # this is assignment, not redef.
   ...
```
To solve this, I detect the newly bound names via AST walking; if rebinds/redefine happen, then I skip the check (entirely in the snippet). Another option could be just raise mypy warning, but then it would flood in a lot of repeated messages that agents can't fix.

2. There are some unresolved symbols that are built into environments. 

I subtract frozenset(n for n in `dir(builtins)` from the reads (mypy lookup). Read resolution itself goes through symtable.is_global, so scope is respected (a nested local/param would not satisfies a module-level read).

3. Handling `__future__` or deferred annotation in general.

In this case mypy resolves annotation names but the interpreter never evaluates a string forward-ref (x: "Foo") or any annotation under from __future__ import annotations, so an unresolved one would raise in mypy while the code runs fine. 

So _has_deferred_annotation detects them (checking prior snippets too for an active future-import) and declines the snippet. Literal[...] members and Annotated[T, "meta"] metadata are not forward-refs, so the detector skips them; only the type position counts to avoid over-declining those common annotations.

### DIFF
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -47,6 +47,12 @@ type ToolCallID = str
 # and it can never collide with a lexical name.
 TYPE_CHECK_ANCHOR_KEY = "<type_check_anchor>"
 
+# REPL `exec_code` snippets ride under their own key so the Callable decoder (which
+# reads TYPE_CHECK_ANCHOR_KEY) never sees a REPL anchor in the shared tool-argument
+# context — preserving the "anchor absent for tool-args" invariant that contracts a
+# synthesized Callable tool argument by its param type, not the Template's return type.
+REPL_ANCHOR_KEY = "<repl_anchor>"
+
 CONTENT_BLOCK_TYPES: frozenset[str] = frozenset(
     literal
     for member in typing.get_args(OpenAIMessageContentListBlock)
@@ -241,7 +247,7 @@ def _pydantic_type_code(ty):
         # the Template body and check it. A type error raises here -> the tool-call decode
         # fails -> `RetryLLMHandler` retries, so ill-typed code never reaches `runcode`.
         ctx = info.context or {}
-        anchor = ctx.get(TYPE_CHECK_ANCHOR_KEY)
+        anchor = ctx.get(REPL_ANCHOR_KEY)
         if anchor is not None:
             # Pass an empty env (not `ctx`): the managed session ignores it, and a fresh
             # fallback session must not be seeded from the decode context (which holds tool
--- a/effectful/handlers/llm/encoding.py
+++ b/effectful/handlers/llm/encoding.py
@@ -216,7 +216,7 @@ def _pydantic_type_code(ty):
     `linecache`, which carries everything the source string did.
     """
 
-    def validate(value: object) -> types.CodeType:
+    def validate(value: object, info: pydantic.ValidationInfo) -> types.CodeType:
         if isinstance(value, types.CodeType):
             return value
         if not isinstance(value, str):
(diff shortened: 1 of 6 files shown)


eb8680:
> A simple solution would be disabling no-redef, but then mypy would raise in case of something like this:

Not sure I quite follow points 1 and 2. My suggestion for the semantics of type-checking for REPL code would be to splice the cumulative code in a given `ReplSession` into the enclosing `Template` body and check the entire module, as you implemented in #706 for `Callable`, with some tweaked mypy settings to prevent spurious errors (e.g. allowing redefinition and missing return statements). That way the code is checked in its execution context.

> Handling __future__ or deferred annotation in general.

`__future__` and star-imports should just be syntactically forbidden in generated code, it's easy and safe to enforce that. Also I think if you follow the previous suggestion to check the code in the `Template` body context you shouldn't run into the issue you described, as all names should be resolved in valid code.

effectful/handlers/llm/evaluation.py
```diff
@@ -554,6 +668,29 @@ def exec_code(self, code: CodeType) -> str:
...
+        # cell may rebind/redefine names and need not return the Template's type.
+        snippet = "".join(linecache.getlines(code.co_filename))
+        if self._anchor is not None:
+            # `_splice_repl` raises on a broken type-check context (source drift, or an
+            # unrecoverable Template anchor) -- let it propagate rather than silently run
+            # unchecked. It returns None only for an empty snippet (nothing to check).
+            checked = _splice_repl(self._prior_snippets, snippet, self._anchor)
+            if checked is not None:
+                try:
+                    type_check(*checked, lenient=True)
```
eb8680: As in `Callable` synthesis, type checking of generated REPL code should happen during decoding of a response into a `types.CodeType` object prior to calling `runcode`. I don't think you should need to change `runcode` or the rest of `ReplSession` at all to achieve the goal of this PR.

effectful/handlers/llm/encoding.py
```diff
@@ -225,9 +225,15 @@ def validate(value: object) -> types.CodeType:
...
-            return evaluation.compile(evaluation.parse(value, filename), filename)
-        except (SyntaxError, ValueError) as exc:
-            raise ValueError(f"source does not compile: {exc}") from exc
+            module = evaluation.parse(value, filename)
+            # Reject `from __future__ import ...` and star imports here, at decode: they
+            # are `SyntaxError` once nested in a function body, so code containing them
+            # can't be spliced into the Template body for type checking. Rejecting at the
+            # decode/compile gate keeps such code from ever becoming a runnable snippet.
+            evaluation.scan_non_nestable(module)
+            return evaluation.compile(module, filename)
```
eb8680: Type checking should happen here, as it already does with `Callable`, rather than inside `ReplSession.runcode`.

effectful/handlers/llm/evaluation.py
datvo06: I moved the REPL type checking to decode as well. I still need to add `_prior_snippets` cause otherwise we would need to read from the temp file where the code is written down many times.

effectful/handlers/llm/encoding.py
datvo06: Updated. I moved it here.

effectful/handlers/llm/evaluation.py
```diff
@@ -365,16 +365,22 @@ def _mypy_check_region(
...
+    # GB (OOM). A subprocess reclaims all of it on exit. Pass a file (not --command:
+    # it hits an argv length limit on large modules); each call gets an isolated temp
+    # dir + cache so parallel decodes don't share -- and deadlock on -- mypy's cache.
     tmpdir = tempfile.mkdtemp(prefix="effectful_typecheck_")
     try:
         tf_path = os.path.join(tmpdir, "_synthesized.py")
         with open(tf_path, "w", encoding="utf-8") as f:
             f.write(source)
-        stdout, stderr, status = mypy_api.run(
+        proc = subprocess.run(
```
datvo06, line 379: Our full real check is more memory-consuming than the previous stub-based type checking. So it is more memory-consuming and thus CI failed. I changed to subprocess-based type checking

eb8680 (approved):

PR #712, opened 2026-07-15
Pin litellm off the wheel-less 1.92.0 (fixes CI on Python 3.14)

### LINKED ISSUE #709: Litellm is incompatible with Python 3.14
It's always had an upper bound (apparently `uv` ignores these?) but now its build breaks on 3.14.

datvo06:
The Python 3.14 matrix job can't install the latest litellm `1.92.0`, so all CIs should fail on all branches.

litellm 1.92.0 publishes no wheel, so `uv` builds it from sdist. On 3.14, that build compiles `pyo3-ffi 0.23.5`, which doesn't support 3.14, so `uv sync` fails at install; fail-fast then cancels the rest of the matrix, so every job goes red before a test runs.

I exclude 1.92.0, so a wheel'd release resolves. Kept the pin in pyproject (not committing uv.lock).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,9 @@ numpyro = [
   "jax<0.10"
 ]
 llm = [
-  "litellm",
+  # 1.92.0 ships no wheel, so it builds from sdist; on Python 3.14 that build pulls
+  # pyo3-ffi 0.23.5, which doesn't support 3.14, so `uv sync` fails the whole matrix.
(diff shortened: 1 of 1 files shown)


datvo06:
Detected this while working on #711.

datvo06:
Also, I'm just skipping this version instead of pinning us permanently below 1.92.0 because litellm might push out whl in the future for different versions.

eb8680 (approved):

PR #713, opened 2026-07-16
Make `__signature__` a cached property

jfeser:
Before:

```
----------------------------------------------------- benchmark: 1 tests -----------------------------------------------------
Name (time in ms)                     Min       Max      Mean   StdDev    Median     IQR  Outliers     OPS  Rounds  Iterations
------------------------------------------------------------------------------------------------------------------------------
test_bench_term_construction     106.4417  171.0169  114.0333  21.3720  107.0401  0.6609       1;1  8.7694       9           1
------------------------------------------------------------------------------------------------------------------------------
```

After:

```
---------------------------------------------------- benchmark: 1 tests ----------------------------------------------------
Name (time in ms)                    Min       Max     Mean   StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------
test_bench_term_construction     60.1234  127.2249  65.1138  16.5825  60.9053  0.8398       1;2  15.3577      16           1
----------------------------------------------------------------------------------------------------------------------------
```

PR #716, opened 2026-07-17
Add cached interpretations and use for `typeof`

jfeser:
Adds a notion of a cached interpretation (these interpretations handle the internal operation `_is_memoized`). When evaluating a term under a cached interpretation, we set a cache attribute on the term that we can reuse for later evaluations under the same interpretation.

Blocked by #713 (only for the benchmark)

Before:
```
---------------------------------------------------- benchmark: 1 tests ----------------------------------------------------
Name (time in ms)                    Min       Max     Mean   StdDev   Median     IQR  Outliers      OPS  Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------
test_bench_term_construction     60.1234  127.2249  65.1138  16.5825  60.9053  0.8398       1;2  15.3577      16           1
----------------------------------------------------------------------------------------------------------------------------
```

After:
```
-------------------------------------------------- benchmark: 1 tests -------------------------------------------------
Name (time in ms)                   Min     Max    Mean  StdDev  Median     IQR  Outliers       OPS  Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------
test_bench_term_construction     4.1599  5.0849  4.4263  0.1847  4.3989  0.2841      35;1  225.9233     111           1
-----------------------------------------------------------------------------------------------------------------------
```

Open design questions:
- The cache needs to distinguish non-equivalent interpretations. It does this with object identity. We could use a stronger comparison operation and get more cache hits. For example, `coproduct({}, i) != i` in the current implementation.
- Whether an interpretation is cached is determined by whether it handles an operation. This means that coproducts with this interpretation will also be cached. Do we want cachability to inherit in this way?

Previous attempt: https://github.com/BasisResearch/effectful/pull/252


jfeser:
#594 is related; it adds a cache that persists for a single evaluate call. This approach is orthogonal. This PR does not improve the evaluation of terms with sharing, and #594 does not reduce the complexity of `defdata`.

effectful/ops/semantics.py
```diff
@@ -20,6 +24,23 @@
 apply = Operation.__apply__
 
 
+@defop
+def _is_memoized() -> bool:
+    """Return whether evaluation under the current interpretation is memoized."""
+    return False
+
+
+def _memoize(intp: Interpretation) -> Interpretation:
```
eb8680: Rather than being able to `_memoize` any `Interpretation` like this, I'd rather define a special `ObjectInterpretation` subtype `PureInterpretation` that is definitionally compatible with caching and restrict this behavior to instances of that. That would also let you remove `_is_memoized`.

effectful/ops/semantics.py
```diff
@@ -166,11 +187,47 @@ def _evaluate_object[T](expr: T, **kwargs) -> T:
...
+    """Return the current memoized interpretation's object identity."""
+    from effectful.internals.runtime import get_interpretation
+
+    intp = get_interpretation()
+    if _is_memoized not in intp:
+        return None
+    return frozenset(intp.items())
+
+
+def _term_cache(expr: Term) -> dict[object, object] | None:
```
eb8680: Would it be cleaner to define this as a `functools.cached_property` of `Term`?

eb8680 (commented):
Looks like a nice improvement, I just have a couple of small comments. 

Also, what happened to removing `internals.product_n`?

effectful/ops/semantics.py
eb8680: Also, if you make `PureInterpretation` hashable, you can make the cache into a `weakref.WeakKeyDictionary`

jfeser: Sure, that sounds reasonable.

effectful/ops/semantics.py
jfeser: This version avoids the need to update implementers of `Term`, but a cached property is cleaner.

jfeser:
I found new uses for `product_n` in #719.

effectful/ops/semantics.py
jfeser: I think since we have a variety of internal and external implementers of `Term`, some of which don't subclass it, we should stick with the current implementation.

effectful/ops/types.py
```diff
@@ -590,6 +591,65 @@ def __call__[**Q, V](
...
+
+    def __iter__(self):
+        raise NotImplementedError
+
+    def __len__(self) -> int:
+        raise NotImplementedError
+
+
+@runtime_checkable
+class PureInterpretation[T, V](Interpretation[T, V], typing.Protocol):
```
eb8680: Not sure this should be a `Protocol`, or even part of the public API for now.

effectful/ops/syntax.py
```diff
@@ -852,6 +837,40 @@ def _(x: object, other) -> bool:
     return x == other
 
 
+class _PureInterpretation[T, V](collections.abc.Mapping):
```
eb8680: Can we just make `PureInterpretation` a subclass of `ObjectInterpretation`? That way we can get rid of `assume_pure` and the `PureInterpretation` `Protocol`.

jfeser: Sure.

effectful/ops/semantics.py
```diff
@@ -248,6 +288,19 @@ def _simple_type(tp: type) -> type:
...
+    @implements(apply)
+    def _(self, op, *args, **kwargs):
+        from effectful.internals.unification import Box
+
+        return Box(op.__type_rule__(*args, **kwargs))
+
+
+def _typeof(term: Expr):
+    """Evaluate the cached type analysis without unwrapping its result."""
+    return evaluate(term, intp=_TypeofIntp())
```
eb8680: Shouldn't you need to use the same instance of `intp` for caching to work, especially with the new `WeakKeyDictionary` cache which should drop the entry for `_TypeofIntp()` after it is GCed? I'd be worried about cache collisions between instances and also memory leaks, especially if this version is somehow managing to cache `typeof` across ephemeral `_TypeofIntp` instances.

effectful/ops/syntax.py
```diff
@@ -929,6 +911,16 @@ def __getitem__(self, item: Operation[..., T]) -> Callable[..., V]:
         return self.implementations[item].__get__(self, type(self))
 
 
+class PureInterpretation[T, V](ObjectInterpretation[T, V]):
+    def __hash__(self):
```
eb8680, line 915: I wonder if this should be more conservative and use object identity instead of a semantic hash of `self.implementations`?

effectful/ops/semantics.py
jfeser: Yes.

effectful/ops/syntax.py
jfeser: I think the semantic hash is correct and slightly more precise, but it doesn't make much difference in our use.

eb8680 (approved):

Issue #664, opened 2026-05-25
Supporting Refinement Type in conjunction with Effect type for stronger security emphasis


datvo06:
Following up on [this comment in #448](https://github.com/BasisResearch/effectful/issues/448#issuecomment-4531026245) - which lays out the motivating LLM-based KG extraction use case, the composition pattern where an edit budget threads through as a State effect, and adjacent instances (taint analysis, capability discipline, validation) - this issue is the concrete proposal for the companion annotation to `Uses(...)` that could be helpful towards security guarantee in modeling in #490 .

`Uses(X)` is a postcondition declaring what an op's handler may perform. `Requires(X)` would be a precondition refining acceptable inputs to "values whose dataflow provenance contains `X`":

```python
@defop
def make_triplet(
    src:      Annotated[Span, Requires(find_span)],
    tgt:      Annotated[Span, Requires(find_span)],
    rel:      str,
    evidence: Annotated[Span, Requires(find_span)],
) -> Triplet: ...
```

Type-theoretically the two aren't symmetric as `Uses` is a classical effect annotation while `Requires` is a refinement on the effect row carried by the value, but they fit the same `Annotation` framework (#477).

### Proposal

1. Add `Requires` as an `Annotation` subtype valid on operation argument positions.
2. Extend `infer_annotations` (#477) to validate `Requires` at op construction and propagate effective effect rows along operation arguments.
3. Specify the propagation rule: an op application's output effective row is the union of its arguments' rows ∪ the op's `Uses` (or analog) set; an argument satisfies `Requires(X)` iff `X` $\in$ `effective_row(arg)`.
4. Provide a public hook so downstream verifiers (provenance, taint, capability) can read a `Term`'s effective row without each one rewalking the DAG.

The `Uses[T]` polymorphism via `Operation` subtyping + `unify` (#349) extends symmetrically `Requires[Tool]` matches subclasses; `Requires[T]` matches via `unify` for the generic case.

### Non-goals (for this issue)

- pyright / static-typechecker integration. The runtime walk over `Term`s is sufficient for v1 downstream.
- Hindley-Milner-style row inference. Annotations are explicit.
- Modifying `Uses(...)`. `Requires` is complementary, not a replacement.

### Relationship to #448

#448 lists "effect types for arguments" as a non-goal. The comment thread there made the case for revisiting it; this issue is the concrete proposal. Happy to fold back into #448 if maintainers prefer.

datvo06:
[...]

PR #717, opened 2026-07-19
Dynamic Effect Type Inference

### LINKED ISSUE #448: Support effect type annotations

datvo06:
Aims to close #448 and #664.

As noted in some of our discussions, static effect typing is pretty impossible given how dynamic we are in selecting handlers and dispatching effects. So this PR adds a dynamic effect type inference. Even though it is dynamic, it should be able to do what we intended, specifically requiring or restricting side effects.

It does this by folding over `apply`: we install an interpretation of apply that records which ops fire, then evaluate the term under it to get the term's effect row: the set of operations it performs. 
To help with this process, I created `usesof` (sibling of `fvsof` and `typeof`) which records the effect set.

Concretely the PR adds:
- `usesof(term) (aliased effectsof)`: the effect row of a term.
- `effect_type(term)`: the pair $\langle \tau, \epsilon \rangle$ of the result type from `typeof` and the effect row from `usesof`, folded over the same apply.
- `Uses[...]` on a return type, read by a new `@final Operation.__uses_rule__` method, plus `check_uses(op, body)`, which returns the effects body performs that the op's declared Uses[...] does not cover. This is the #448 checker: empty means the declaration holds, and it is transitively closed because `usesof` unions the whole term.
- `Computation` on a callback argument, so a higher-order op folds the effects hidden inside its callback instead of the checker hard-coding which ops are higher-order.
- `Requires(...)` on an argument plus check_requires(term) (this is from #664 but I think this is quite small so I left it in): an argument satisfies `Requires(X)` iff `X` is in `usesof(arg)`.
- `toolsof / reachable_tools / check_tools` in `handlers/llm`: the static tool graph an LLM template can reach, computed by reifying to a term and reading `.tools`, with no LLM call. `check_tools(fn, *allowed)` is the tool-graph analog of `check_uses`.


datvo06:
A few honest caveats, since dynamic inference is not free:

- `usesof` is path insensitive over the reified term, so it is an over-approximation and it needs the term to be straight-line: native Python control flow gets resolved when the term is built, so a branchy function can under-approximate. The same precondition applies to `reachable_tools`.
- The `Computation` callback runs the callback on an opaque placeholder and raises an error (`UnsoundCallbackFold`) when the callback inspects that placeholder through the operator protocol (`==`, arithmetic, len, calling it, indexing). But object identity (`is`), `type()` and `isinstance` bypass the protocol and it cannot catch them, so a callback branching on those under-approximates silently. I've put these edge cases into the tests.

datvo06:
I deferred the following LLM-specific features pending on #694 :
- The LLM tool restriction from `#448` (only offering a template the tools it lists) belongs on the tool-assembl, not on `Template.tools`, so it lands separately once that seam is in place, I think after #694 landed, I can wire these in easily.
- `Uses` members that are `Operation` subclasses or string self-references, the build-time validation of annotations at op construction, and the usage multiset (usagesof) plus handler discharge.

datvo06:
I checked most of the logics here, but I think I can review more until marking it ready for review.

eb8680:
This looks interesting, but I think the specification of the intended behavior from #448 and #664 needs to be a lot clearer and a lot more grounded in relevant PL literature and the semantics of `effectful` before we can actually implement anything (or trust anything cooked up by Claude).

I also think there are many outstanding issues in `effectful` and downstream that are higher priority, especially given that lexical scoping already solves closely related problems (#590) and the large amount of work this will inevitably need before it's useable in practice.

datvo06:
> This looks interesting, but I think the specification of the intended behavior from #448 and #664 needs to be a lot clearer and a lot more grounded in relevant PL literature and the semantics of `effectful` before we can actually implement anything.

Got it. I ended up at this after failing to find a good relevant paper and then designing it based on stateless (we use `apply` instead of `yield`). But I will keep looking at PL literature to see if something relevant come up.

Will also work on #590!

PR #719, opened 2026-07-22
Add collection constructor operations and rewrite analyses to remove side effects

### LINKED ISSUE #715: `evaluate` should call constructor operations when rebuilding collections
This is a proposal to make evaluation of terms that contain collections more uniform and easier to control using handlers.

We should modify `evaluate` so that it calls a construction operation when evaluating through a collection. This would simplify the implementation of any interpretation that provides a non-standard evaluation that doesn't preserve collection structure.

For example, `evaluate([a(), {0: (b(), c()}])` should call `as_tuple`, `as_dict`, and `as_list` (effectful wrappers of the stdlib collection constructors) when rebuilding the respective collections. An analysis that returns the set of nullary operations in a term could then be written as:

```py
class Vars(ObjectInterpretation):
    @implements(apply)
    def _(op, *args, **kwargs):
        return {op} if not args and not kwargs else set().union(*args, *kwargs.values())
```

Currently, analyses like this need to implement their own collection flattening or use mutable state (see `fvsof`, `sizesof`).

We can retain the existing term representation that allows embedded collections, but this change would make terms truly equivalent to trees of operations applied to values.

https://github.com/BasisResearch/effectful/pull/641 is an ad-hoc version of this change that only treats dataclass constructors.

jfeser:
Closes #715, #624 

I did not pick up the changes from #680, although they're of a similar flavor.

Rewrites `fvsof` and `sizesof` in the jax module to remove their side-effects. This makes them compatible with memoization from #716.

~~`fvsof`'s behavior changes for terms that contain collections: the collection construction operations now appear in the term free variables.~~

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -231,7 +231,7 @@ def _evaluate_dataclass[T](expr: T, **kwargs) -> T:
     dataclass_op = _get_dataclass_constr_op(type(expr))
     subst = {
         field.name: evaluate(getattr(expr, field.name))
-        for field in dataclasses.fields(expr)
+        for field in dataclasses.fields(expr)  # type: ignore[arg-type]
     }
     return typing.cast(T, dataclass_op(**subst))
 
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -211,17 +211,29 @@ def evaluate[T](
 @evaluate.register(bytes)
 def _evaluate_object[T](expr: T, **kwargs) -> T:
     if dataclasses.is_dataclass(expr) and not isinstance(expr, type):
-        return typing.cast(
-            T,
-            dataclasses.replace(
-                expr,
-                **{
-                    field.name: evaluate(getattr(expr, field.name))
-                    for field in dataclasses.fields(expr)
-                },
-            ),
-        )
+        return _evaluate_dataclass(expr, **kwargs)
     return expr
+
+
+def _get_dataclass_constr_op(typ):
+    if hasattr(typ, "constr_op"):
+        return typ.constr_op
+
+    @Operation.define
+    def constr_op(*args, **kwargs) -> typ:
+        return typ(*args, **kwargs)
+
+    typ.constr_op = constr_op
+    return constr_op
+
+
+def _evaluate_dataclass[T](expr: T, **kwargs) -> T:
+    dataclass_op = _get_dataclass_constr_op(type(expr))
+    subst = {
+        field.name: evaluate(getattr(expr, field.name))
+        for field in dataclasses.fields(expr)
+    }
+    return typing.cast(T, dataclass_op(**subst))
 
 
 @evaluate.register(Term)
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -360,7 +360,7 @@ def fvsof[S](term: Expr[S]) -> collections.abc.Set[Operation]:
         # arguments are available through argsof while this node is analyzed.
         return _BaseTerm(op)
 
-    def _apply_collection_fvs(op, *args, **kwargs):
+    def _apply_passthrough_fvs(op, *args, **kwargs):
         return frozenset().union(
             *(x for x in (*args, *kwargs.values()) if isinstance(x, frozenset))
         )
@@ -394,7 +394,8 @@ def fvsof[S](term: Expr[S]) -> collections.abc.Set[Operation]:
         {
             _fvsof_fvs: {
                 apply: _apply_fvs,
-                CollectionConstrOperation.__apply__: _apply_collection_fvs,
+                CollectionConstrOperation.__apply__: _apply_passthrough_fvs,
+                DataclassConstrOperation.__apply__: _apply_passthrough_fvs,
             },
             _fvsof_binders: {
                 CollectionConstrOperation.__apply__: _apply_collection_binders,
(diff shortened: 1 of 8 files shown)


eb8680 (commented):
> fvsof's behavior changes for terms that contain collections: the collection construction operations now appear in the term free variables.

This seems like a big breaking change, are we sure it's necessary? I would expect lots of places here and downstream to depend on tests like `if not fvsof(x): ...` to gate partial evaluation behavior.

jfeser:
No, it's not, but we should think about what semantics we want for `fvsof`. One way to view this change is that collections are a kind of shorthand for the real term, which is exactly an operation tree; in this view, `fvsof` should reflect that. Another view is that collection operations are an internal convenience that we want to hide from users, and should also be hidden from `fvsof`. I think the second view is the pragmatic one. Both are easy enough to implement.

jfeser:
This PR now doesn't change the behavior of `fvsof`.

effectful/ops/semantics.py
```diff
@@ -115,6 +123,18 @@ def handler(intp: Interpretation):
         yield intp
 
 
+class DataclassConstrOperation(Operation): ...
```
eb8680: Should share a common base with `CollectionConstrOperation`

effectful/ops/syntax.py
```diff
@@ -1327,3 +1338,26 @@ class _IntegralTerm[T: numbers.Integral](_RationalTerm[T]):
 @defdata.register(bool)
 class _BoolTerm[T: bool](_IntegralTerm[T]):  # type: ignore
     pass
+
+
+class CollectionConstrOperation(Operation): ...
+
+
+@CollectionConstrOperation.define
+def as_tuple(*args) -> tuple:
```
eb8680: This type rule is wrong/too weak, I would expect this to break `typeof` for tuples

effectful/ops/syntax.py
```diff
@@ -1327,3 +1338,26 @@ class _IntegralTerm[T: numbers.Integral](_RationalTerm[T]):
 @defdata.register(bool)
 class _BoolTerm[T: bool](_IntegralTerm[T]):  # type: ignore
     pass
+
+
+class CollectionConstrOperation(Operation): ...
```
eb8680: I think this needs at least one type parameter for the return type

effectful/ops/semantics.py
```diff
@@ -115,6 +123,18 @@ def handler(intp: Interpretation):
...
+
+
+@functools.cache
+def _as_type(typ: type, operation_type=CollectionConstrOperation):
```
eb8680: I wonder if this should be a public `classmethod` of one of the new `Operation` classes. It seems like the right way to access these operations is to look them up by return type rather than by name, and to have an extension API that allows someone to register new ones as with to the `singledispatch`-based extensibility of `evaluate` and `defdata`.

effectful/ops/syntax.py
```diff
@@ -1327,3 +1338,26 @@ class _IntegralTerm[T: numbers.Integral](_RationalTerm[T]):
...
+
+
+@CollectionConstrOperation.define
+def as_tuple(*args) -> tuple:
+    return tuple(args)
+
+
+@CollectionConstrOperation.define
+def as_list[T](*args: T) -> list[T]:
```
eb8680: As I wrote in another comment, I think we probably want to look these up by type (in a `singledispatch` registry that is subtyping-aware) rather than by name to preserve our extensibility story

eb8680 (commented):
How does this affect extensibility? With `evaluate`, `defdata`, `nested_type` etc the story is clear - a downstream user can register type-specific implementations that make them compatible with new third-party types. We don't want this new mechanism to break that and require edits to `effectful` to support new types. 

I'd also worry that this is proliferating rather than consolidating traversal logic (#682 ) - isn't this morally duplicating `evaluate`'s type-unpacking behavior? How do we avoid that, or at least enforce consistency across traversals?

effectful/ops/syntax.py
jfeser: `typeof` runs `CollectionConstrOperation`s rather than handling them. This allows us to keep the smarter type checking behavior from `nested_type`. I don't think we have rich enough types at the moment to type check the tuple constructor.

jfeser:
`evaluate` extensions would need to use `_as_type` to allow uniform handling by analyses. This requires updating the extenders of `evaluate`, but I think it's better than the status quo.

Right now, `typeof` works because both `evaluate` and `nested_type` can be extended; `nested_type` unpacks the collection-typed analysis results produced by `evaluate`. An extension to `evaluate` generally requires an extension to `nested_type`. Some analyses, like `fvsof`, don't have the equivalent of `nested_type`, but can be implemented correctly using side effects. Other analyses, like `Term.__str__`, simply don't work correctly in the presence of extensions.

This PR doesn't change the status quo around `nested_type`, mainly because I think that's a substantial change, and I'm not sure we actually have rich enough type annotations to capture what `nested_type` does.

Re, traversal logic, I think this is a simplification because more analyses can become pure handlers of `apply` (as is done in this PR for `fvsof` and `sizesof`). I'm not totally sure what it would be mean for interpretations to be consistent about their traversal behavior beyond consuming the term tree in the same order.

effectful/ops/semantics.py
jfeser: Sure.

effectful/ops/syntax.py
jfeser: These operations aren't the same as their corresponding constructors (e.g. `tuple` takes a single sequence argument, not a sequence of arguments). I think we only need a single sequence constructor though (probably `as_tuple`).

eb8680 (approved):

PR #722, opened 2026-07-23
Change pytest-xdist flag in CI from 'auto' to 'logical'

eb8680:
Following a similar PR in RoboTL, this PR is an attempt to speed up CI further by parallelizing across logical CPUs


eb8680:
It does seem to be quite a bit faster thanks to the extra parallelism for everything except the `numpyro` tests, which are unfortunately the slowest ones. Still definitely worth merging, though.

jfeser (approved):

PR #724, opened 2026-07-25
Add comprehension syntax for `Monoid`

eb8680:
This PR adds a `Monoid.__call__` method and a whole bunch of internal machinery based on #288 to allow the use of generator comprehension syntax with `Monoid`s under very general conditions.


effectful/ops/monoid.py
```diff
@@ -120,6 +120,30 @@ def __eq__(self, other):
     def __hash__(self):
         return hash(id(self))
 
+    def __call__(self, comprehension: Generator[Any, None, None]) -> Expr[W]:
```
eb8680: This new `__call__` method on `Monoid` is the only public-facing change in this entire PR. Everything else in the diff is solely there to implement this interface.

eb8680:
Closing in favor of constituent PRs

PR #726, opened 2026-07-28
Fix quadratic term construction under nested binders

eb8680:
## The regression

#716 ("Add cached interpretations and use for `typeof`") made `defdata` compute the
full type analysis of every term it builds, in order to pick the right term
constructor. That interacts badly with binders: when an operation binds variables,
`defdata` renames them throughout the body, which rebuilds the body by re-entering
`defdata` via `apply`. Re-entering `defdata` recomputes binders and renames each
child *again* at every level, so nesting binders re-traverses the subtree once per
level of nesting.

## The changes

- **`effectful/ops/syntax.py`**: `evaluate_with_renaming` now rebuilds nodes with
  `_build_term` instead of re-entering `defdata`. The subterm was already renamed
  when it was first built, so the rebuild only needs to re-dispatch each node, not
  redo binder analysis and renaming.
- `_build_term` is `defdata` minus the renaming step, and is shared: `defdata`
  itself ends with `return _build_term(__dispatch, op, *args_, **kwargs_)`, and the
  rebuild installs `functools.partial(_build_term, __dispatch)` as the `apply` rule.

The change does not alter any observable result.

## Benchmark

`tests/test_ops_syntax.py::test_bench_nested_binder_construction` builds a term
of 10 nested `let`-style binders, each variable used in the body below it. The
existing `test_bench_term_construction` builds a *binder-free* term, so it never
reaches this path and stays fast even when nested construction blows up.

Measured on this branch vs. the same test with the `effectful/` change reverted
(mean of pytest-benchmark rounds):

| benchmark | before | after |
| --- | --- | --- |
| `test_bench_nested_binder_construction` | 572.8 ms | 17.2 ms (**33x**) |
| `test_bench_term_construction` | 3.39 ms | 3.50 ms (unchanged) |

An earlier revision of this PR also seeded `defdata`'s already-computed type
analysis into the per-term evaluation cache, which bought a further ~1.8x on
nested construction and ~1.5x on binder-free construction. That was dropped at
review: it reached past `_evaluate_term`, which owns that cache, and it is a
constant factor rather than the asymptotic fix (with and without, nested
construction is quadratic in depth, at a flat ~1.8x ratio from depth 10 to 40).

## Testing

Full test suite: 18749 passed, 1 skipped, 2108 xfailed. `mypy` and `ruff` clean
on the changed files.

Split out of #724 for review.

### DIFF
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -306,6 +306,25 @@ class _TypeofIntp(PureInterpretation):
 _TYPEOF_INTP = _TypeofIntp()
 
 
+def _seed_typeof(expr: Expr, full_type: Any) -> None:
+    """Record a term's already-computed type analysis in its cache.
+
+    :func:`~effectful.ops.syntax.defdata` computes the type of every term it
+    builds in order to pick a constructor. Storing that result here means a
+    parent's :func:`_typeof` on a freshly built child is a cache hit rather
+    than a fresh traversal of the whole subterm -- without which term
+    construction is quadratic in subterm size, and compounds multiplicatively
+    through nested binders.
+    """
+    from effectful.internals.unification import Box
+
+    if not isinstance(expr, Term):
+        return
+    cache = _term_cache(expr)
+    if cache is not None:
+        cache[_TYPEOF_INTP] = Box(full_type)
+
+
 def _typeof(term: Expr):
     """Evaluate the cached type analysis without unwrapping its result."""
     return evaluate(term, intp=_TYPEOF_INTP)
--- a/effectful/ops/semantics.py
+++ b/effectful/ops/semantics.py
@@ -306,25 +306,6 @@ class _TypeofIntp(PureInterpretation):
 _TYPEOF_INTP = _TypeofIntp()
 
 
-def _seed_typeof(expr: Expr, full_type: Any) -> None:
-    """Record a term's already-computed type analysis in its cache.
-
-    :func:`~effectful.ops.syntax.defdata` computes the type of every term it
-    builds in order to pick a constructor. Storing that result here means a
-    parent's :func:`_typeof` on a freshly built child is a cache hit rather
-    than a fresh traversal of the whole subterm -- without which term
-    construction is quadratic in subterm size, and compounds multiplicatively
-    through nested binders.
-    """
-    from effectful.internals.unification import Box
-
-    if not isinstance(expr, Term):
-        return
-    cache = _term_cache(expr)
-    if cache is not None:
-        cache[_TYPEOF_INTP] = Box(full_type)
-
-
 def _typeof(term: Expr):
     """Evaluate the cached type analysis without unwrapping its result."""
     return evaluate(term, intp=_TYPEOF_INTP)
--- a/effectful/ops/syntax.py
+++ b/effectful/ops/syntax.py
@@ -491,7 +491,13 @@ def defdata[T](
     it is reconstructed as a :class:`_CallableTerm`, which implements the :func:`__call__` method.
     """
     from effectful.internals.runtime import interpreter
-    from effectful.ops.semantics import _simple_type, _typeof, apply, evaluate
+    from effectful.ops.semantics import (
+        _seed_typeof,
+        _simple_type,
+        _typeof,
+        apply,
+        evaluate,
+    )
 
     # If this operation binds variables, we need to rename them in the
     # appropriate parts of the child term.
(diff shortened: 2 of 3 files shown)


effectful/ops/syntax.py
```diff
@@ -514,9 +520,24 @@ def evaluate_with_renaming(expr, ctx):
...
-        with interpreter({apply: defdata} | renaming_ctx):
+        #
+        # Rebuild with ``_reconstruct`` rather than ``defdata``: the subterm was
+        # already renamed when it was first built, so re-entering ``defdata``
+        # here would recompute binders and rename each child again at every
+        # level, re-traversing the subtree once per level of nesting.
+        with interpreter({apply: _reconstruct} | renaming_ctx):
             return evaluate(expr)
 
+    def _reconstruct(op, *args, **kwargs):
```
jfeser: It would be nice to remove the duplication between this rule and the rest of the code in `defdata`. I think they have the same behavior.

effectful/ops/semantics.py
```diff
@@ -306,6 +306,25 @@ def _(self, op, *args, **kwargs):
 _TYPEOF_INTP = _TypeofIntp()
 
 
+def _seed_typeof(expr: Expr, full_type: Any) -> None:
```
jfeser: I don't think we need this for performance, and it violates the abstraction boundary around the cache.

jfeser (changes requested):

effectful/ops/semantics.py
eb8680: Agreed on both counts — removed, along with the whole `_seed_typeof` function. `effectful/ops/semantics.py` is now untouched by this branch.

You're right that it isn't what fixes the regression. `_evaluate_term` already caches a node's typeof result whenever `_typeof` is called on that node, so by the time `defdata` finishes building a node, every descendant is already cached and the parent's `_typeof` on the fresh child is one level deep, not a subtree walk. The asymptotic fix is entirely the `apply` change, which stops the rebuild from re-entering `defdata` and re-running `__fvs_rule__` + renaming at every level.

I measured before dropping it, since the description had claimed otherwise:

| `test_bench_nested_binder_construction` | mean |
| --- | --- |
| before this branch | 572.8 ms |
| this branch, `_seed_typeof` removed | 17.2 ms (**33x**) |
| this branch as originally posted | 9.4 ms (61x) |

And a scaling check, since a 1.8x gap could have hidden an order-of-growth difference — nested-binder construction, min-of-runs:

| depth | without `_seed_typeof` | with |
| --- | --- | --- |
| 10 | 14.2 ms | 8.5 ms |
| 20 | 56.1 ms | 31.3 ms |
| 30 | 123.1 ms | 67.1 ms |
| 40 | 217.3 ms | 116.7 ms |

Both are quadratic in depth (4x depth → ~14x time in both), with a flat 1.67/1.79/1.84/1.86 ratio. So it was a constant factor, not a scaling fix — worth ~1.8x on nested construction and ~1.5x on binder-free construction, which this branch now gives up. Flagging that explicitly so it reads as a deliberate trade rather than a silent regression: if that constant turns out to matter, the right shape is a sanctioned "record a known result" entry point next to `_term_cache` in `semantics.py`, so `syntax.py` never touches the cache dict or knows that typeof results are `Box`-wrapped. Happy to do that separately if you want it.

effectful/ops/syntax.py
eb8680: They did, and once `_seed_typeof` was gone they were identical. Both are now the module-level `_build_term`, which is `defdata` minus the renaming step:

```python
def _build_term[T](
    __dispatch: Callable[[type], Callable[..., Expr[T]]],
    op: Operation[..., T],
    *args,
    **kwargs,
) -> Expr[T]:
    from effectful.ops.semantics import _simple_type, _typeof

    typed_args = tuple(_typeof(arg) for arg in args)
    typed_kwargs = {k: _typeof(v) for k, v in kwargs.items()}
    dispatch_type = _simple_type(op.__type_rule__(*typed_args, **typed_kwargs))
    return __dispatch(dispatch_type)(dispatch_type, op, *args, **kwargs)
```

`defdata` ends with `return _build_term(__dispatch, op, *args_, **kwargs_)`, and the renaming rebuild installs `functools.partial(_build_term, __dispatch)` as the `apply` rule. It takes `__dispatch` as a parameter rather than closing over it so it can live at module level instead of being rebuilt on every `defdata` call; the `partial` is only allocated on the renaming path, which the binder-free path skips at `if not renaming_ctx: return expr`. Moving the deferred `semantics` import from once-per-`defdata` to once-per-node costs nothing measurable (depth 14: 27.99 ms before the refactor, 27.1 ms after, min-of-runs).

jfeser (approved):
Fine to merge once CI passes.

PR #742, opened 2026-07-31
Generalize `ReduceEqualityMaskRange`

jfeser:


effectful/ops/monoid.py
```diff
@@ -420,6 +449,49 @@ def _(self, monoid, body, streams):
         return fwd()
 
 
+class ReduceIntersectionSingletonRange(ObjectInterpretation):
```
eb8680, line 626: This seems like a very specific pattern, when does this arise?

jfeser: This is the pattern produced by the less general equality mask elimination rule.

effectful/ops/monoid.py
```diff
@@ -390,95 +443,226 @@ def group_plus(value):
...
-        for i, cond in enumerate(conds):
-            matched = self._match_eq(cond, streams)
-            if matched is None:
+        matched_conds = [
+            (i, solution.stream_op, solution.rhs)
+            for i, cond in enumerate(conds)
+            for solution in _solve_stream_equality(cond, streams)
+            if cls._can_restrict(streams[solution.stream_op])
+        ]
+        ranked_conds = sorted(matched_conds, key=lambda c: (*cls._rhs_cost(c[2]), c[0]))
```
eb8680, line 570: What is being added by this new cost-based sorting step here? If it's important, is there a way to achieve the same thing through `choose_contraction`, so there's only a single source of cost-based non-determinism? If it's not important, can we just retain the original order in `matched_conds` and remove it?

effectful/ops/semantics.py
```diff
@@ -427,3 +427,37 @@ def fvsof[S](term: Expr[S]) -> collections.abc.Set[Operation]:
...
+            x if isinstance(x, _SizeofAnalysisValue) else _SizeofAnalysisValue(1)
+            for x in (*args, *kwargs.values())
+        ]
+        size = sum(s_args, start=_SizeofAnalysisValue(1))
+        return size
+
+    return {apply: _apply}
+
+
+def sizeof(term: Expr) -> int:
```
eb8680, line 459: I only see one place in `ops.monoid` that uses `sizeof`. If this can't be removed, do you expect it to be used anywhere else in the future? It seems like a very questionable basis for a cost model.

jfeser: This code is still in flux, so I'm not sure. In the robotl use case, effectful overhead is the main cost (both before and after the caching work). A cost model that reduces term growth is essential. I'd expect future rewriting rules to have similar problems with creating huge terms, so I expect `sizeof` to remain useful.

PR #743, opened 2026-08-03
Unify caching mechanisms and separate their management from `Interpretation`s

eb8680:
Addresses #633 

This PR implements a new `weakref`-based caching system that unifies and generalizes those introduced in #726 and #594 

Specifically, it adds a new context manager `effectful.internals.runtime.cache` such that within a `cache()`-delimited block, repeated calls to `evaluate` with the same expression or data structure and active interpretation are cached.

This new mechanism is backed by a new standalone utility module `effectful.internals.weak`, based on a [similar utility module `torch.utils.weak` in PyTorch](https://github.com/pytorch/pytorch/blob/708f706e7ac19247a4d6f26e81c9c706ac14d50d/torch/utils/weak.py), which implements a couple of different cache data structures generalizing `weakref.WeakKeyDictionary`:
- `WeakIdKeyDictionary` behaves like `WeakKeyDictionary` (including forbidding keys which cannot be `weakref.ref`ed, like `tuple`s and `int`s) but uses object identity rather than object equality/hashing to discriminate between keys. (This is the main thing provided by `torch.utils.weak`)
- `IdKeyDictionary`, which stores strong references to keys that it discriminates between using identity
- `AutoIdKeyDictionary`, which extends `WeakIdKeyDictionary` and `IdKeyDictionary`'s identity-based semantics by attempting to use a weak reference for a new key and falling back to a strong reference when that is disallowed by the key type, and ensuring that weak reference and strong reference keys cannot conflict with one another.
- `weak_memoize`, a helper function that can use these data structures to memoize a single-argument function analogous to `functools.cache`

The PR also includes a number of small caching-related semantics-preserving performance fixes, and a benchmark derived from a RoboTL issue illustrating both asymptotic and constant-factor improvements on a representative workload.

### DIFF
--- a/effectful/internals/runtime.py
+++ b/effectful/internals/runtime.py
@@ -132,6 +132,24 @@ def _save_args[**P, T](fn: Callable[P, T]) -> Callable[P, T]:
     return _cont_wrapper
 
 
+@weak_memoize
+def _save_then_restore_args[**P, T](fn: Callable[P, T]) -> Callable[P, T]:
+    # should be equivalent to _restore_args(_save_args(fn)), just fused
+    from effectful.ops.semantics import handler
+
+    sig = inspect.signature(fn)
+    if not sig.parameters:
+        return fn
+
+    @functools.wraps(fn)
+    def _cont_wrapper(*a: P.args, **k: P.kwargs) -> T:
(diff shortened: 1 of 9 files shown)


eb8680:
Splitting into two separate PRs for easier review

PR #744, opened 2026-08-04
Add standalone module `internals.weak` with data structures for caching

eb8680:
Factored out of #743 to ease review.

This PR contains only the new internal-facing module `effectful.internals.weak` and its tests `tests/test_internals_weak.py` from #743. This module and its tests are self-contained and do not depend on any other part of `effectful`; no changes here affect the rest of the library in any way.

Most of the implementation of `internals.weak.WeakIdKeyDictionary` is copied from `torch.utils.weak`, although I've made some fixes and cleaned up types and docstrings. The things on top of that are new.

### DIFF
--- a/effectful/internals/weak.py
+++ b/effectful/internals/weak.py
@@ -101,13 +101,6 @@ class WeakIdRef[T](weakref.ref[T]):
         self._id = id(key)
         super().__init__(key, callback)  # type: ignore[call-arg]
 
-    def __call__(self) -> T | None:
-        r = super().__call__()
-        # Special logic for Tensor PyObject resurrection
-        if r is not None and hasattr(r, "_fix_weakref"):
-            r._fix_weakref()
-        return r
-
     def __hash__(self) -> int:
         return self._id
 
--- a/effectful/internals/weak.py
+++ b/effectful/internals/weak.py
@@ -1,0 +1,639 @@
+"""Dictionaries that key on object identity rather than ``__eq__`` and ``__hash__``.
+
+Adapted from pytorch's ``torch/utils/weak.py`` at 708f706, which is itself
+adapted from cpython's ``Lib/weakref.py``:
+https://github.com/pytorch/pytorch/blob/708f706e7ac19247a4d6f26e81c9c706ac14d50d/torch/utils/weak.py
+
+A stock :class:`weakref.WeakKeyDictionary` compares its keys with ``==`` and
+hashes them, which rules out most of what this package wants to cache on. A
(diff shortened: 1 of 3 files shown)


tests/test_internals_weak.py
```diff
@@ -0,0 +1,1549 @@
+"""Tests for :mod:`effectful.internals.weak`.
+
+Two complementary strategies:
+
+* Hand-written invariant tests, parameterised over the dictionary flavours, that
+  name the law they check (``M3``, ``K5``, ...). These give a named failure.
```
jfeser: These laws aren't defined. What do they refer to?

jfeser (changes requested):
I've pushed a failing test that should pass. I haven't read this code in detail, just put it through GPT.

tests/test_internals_weak.py
eb8680: They're references to a design document I made Claude write. I'll inline the definitions in the test docstrings.

eb8680:
I fixed the test you pushed. 

> I haven't read this code in detail, just put it through GPT.

The nastiest part of the code is the implementation of `WeakIdKeyDictionary` (including `_iterationGuard` and `WeakIdRef`) which is a forked and lightly modified version of `torch.utils.weak.WeakIdKeyDictionary` and thus shouldn't need careful review. The other stuff in `weak.py` is more straightforward and largely sits on top of that. The tests were generated by Claude from a specification I wrote and reviewed with/for it, along with the tests from PyTorch for `torch.utils.weak` (which are not exactly comprehensive). I am never thrilled with Claude's test code but I think it covered everything in the spec.

effectful/internals/weak.py
```diff
@@ -0,0 +1,647 @@
...
+        # cache the id of the key as we know this is definitely the hash
+        # method
+        self._id = id(key)
+        super().__init__(key, callback)  # type: ignore[call-arg]
+
+    def __call__(self) -> T | None:
+        r = super().__call__()
+        # Special logic for Tensor PyObject resurrection
+        if r is not None and hasattr(r, "_fix_weakref"):
+            r._fix_weakref()
```
jfeser: If we keep this tensor-specific code, we should have a test.

jfeser (changes requested):

effectful/internals/weak.py
jfeser: Looks like torch fixed its issues with weakref in pytorch/pytorch#167564.

jfeser (approved):

PR #747, opened 2026-08-05
Remove `internals.product_n` and speed up `fvsof` and `sizesof`

eb8680:
Blocked by #745 

This PR refactors `fvsof`, `handlers.torch.sizesof` and `handlers.jax.sizesof` to make use of the new caching mechanism in #745 in lieu of `productN`, which seems to speed them all up by a constant factor of ~5x without affecting their semantics. It also drops `product_n.py` and `test_internals_product_n.py`, which are now unused. It also shares the new implementation of `sizesof` across the Torch and JAX backends, improving correctness and maintainability.


eb8680:
I think this should have a nice side effect of speeding up the NumPyro CI build, which is the slowest part of our test suite that bottlenecks CI times.

effectful/handlers/jax/_handlers.py
```diff
@@ -224,14 +133,46 @@ def _jax_op(*args, **kwargs) -> jax.Array:
...
     """
     return x[tuple(key)]
 
 
+class _SizesofIntp(_BaseSizesofIntp[jax.Array]):
+    arr_type: typing.ClassVar[type] = jax.Array
+
+    @classmethod
+    def _names_dim(cls, op: Operation[[], jax.Array]) -> bool:
+        return True
```
jfeser, line 149: Why does this logic differ between the torch and jax implementations?

jfeser (changes requested):
`fvsof` shouldn't return renamed bound operations from collection binders.

effectful/handlers/jax/_handlers.py
eb8680: I don't know, that's just the current behavior. I'm guessing it can be made the same by changing `handlers.torch` with no issues, but I wanted the changes in this PR to be fully semantics-preserving.

jfeser: Ok. Let's open an issue as a reminder that we should fix this.

eb8680: #750

eb8680:
Should be fixed now

jfeser (approved):

PR #752, opened 2026-08-07
Add `Group` and associated rules

jfeser:
Adds `Group`s (monoids with inverses) and normalization rules. The first intended user of this code is in the pose solver in RobotL.


eb8680 (approved):
I'll go ahead and merge this, but why stop at `Sum`? `Product` and `LogSumExp` have inverses too.

jfeser:
`Product` has a partial inverse. I wasn't excited about trying to discharge the resulting nonzero constraint. I'm not sure about `LogSumExp`.

PR #756, opened 2026-08-13
Validate a submit_solution answer against its declared return type

datvo06:
In #694, a Template's return-type validator only runs when the model answers with a direct final message. When it answers via the `submit_solution` final tool, the answer is returned unvalidated

This PR runs the declared return type's value-validators on a `submit_solution` answer.


eb8680:
I pushed a commit to #694 that should fix this and the broader issue of `PlainValidator` erasing other validators.

PR #760, opened 2026-08-31
Forwarding proceeds through apply to reach default rule

### LINKED ISSUE #759: Apply handlers should be possible targets of `fwd`
The following test should pass:
```py
from effectful.ops.semantics import apply, fwd, handler
from effectful.ops.types import Operation


@Operation.define
def f(x: int) -> int:
    return x + 1


f_handler_calls = 0
apply_handler_calls = 0


def _f_handler(x):
    global f_handler_calls
    f_handler_calls += 1
    return fwd()


def _apply_handler(op, *args, **kwargs):
    global apply_handler_calls
    apply_handler_calls += 1
    return fwd()


with handler({apply: _apply_handler}), handler({f: _f_handler}):
    assert f(1) == 2
    assert f_handler_calls == 1
    assert apply_handler_calls == 1
```

Apply handlers are often used as catch-all handlers to implement tracing or similar global effects. This doesn't work right now unless the traced effects are otherwise unhandled.

jfeser:
Closes #759 

This changes the behavior of `fwd` so that `__apply__` is called instead of reaching the default rule when forwarding from the last handler in the chain. This gives `__apply__` handlers a chance to see the operation. The operation default rule is now called from within the default rule of `__apply__`.

This behavior does not apply to the `__apply__` operations themselves. They are now instances of `ApplyOperation` to facilitate making this distinction.

### DIFF
--- a/effectful/ops/types.py
+++ b/effectful/ops/types.py
@@ -533,8 +533,7 @@ class Operation[**Q, V]:
             fwd_intp = typing.cast(Interpretation, {fwd: self._next_rule_with_args})
             with handler(fwd_intp):
                 return self_handler(*args, **kwargs)
-        elif args and isinstance(args[0], Operation) and self is args[0].__apply__:
-            # Prevent infinite recursion when calling self.apply directly
+        elif isinstance(self, ApplyOperation):
             return self.__default__(*args, **kwargs)
         else:
             return self.__apply__(self, *args, **kwargs)
@@ -544,9 +543,7 @@ class Operation[**Q, V]:
         if not _generate_apply:
             return
 
-        assert "__apply__" not in cls.__dict__ or cls is Operation, (
-            "Cannot manually override apply"
-        )
+        assert "__apply__" not in cls.__dict__, "Cannot manually override apply"
         assert isinstance(cls.__apply__, ApplyOperation)
 
         cls.__apply__ = cls.__apply__.define(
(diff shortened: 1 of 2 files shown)


eb8680 (commented):
I'm not sure all the special-casing around `ApplyOperation` is right. Why do we need that type at all? Why shouldn't behavior be uniform with the sole exception of the base `Operation.__apply__`?

jfeser:
A version that only special cases `Operation.__apply__` can work as well, but it involves more calls to `__apply__`. For example, the call sequence in `test_fwd_through_apply_operation_subtypes` would become `['exact', 'derived apply', 'apply', 'base apply', 'apply', 'apply', 'default']`.

eb8680:
That test seems especially suspect. Why should calling `fwd` in a handler of `DerivedOperation.__apply__` jump directly up the `DerivedOperation` MRO to a handler of `BaseOperation.__apply__`, as opposed to going to `Operation.__apply__`'s rule for `DerivedOperation.__apply__`? Won't this interfere with the way bound-method operations currently resolve behavior by going instance -> class in their default rule and using `super()` to traverse their owners' MROs? I think we need a diagram or something that clarifies the interactions between the different resolution mechanisms once and for all before we change the status quo behavior.

jfeser:
If we run the same test on `master`, but remove the handler for `f`, we get `calls = ['derived apply', 'base apply', 'apply', 'default']`. This direct forwarding behavior is the current semantics, not a new change.

eb8680 (approved):

eb8680:
I merged this because of the downstream dependency but I would like to resolve the correctness question #783 at some point.

PR #768, opened 2026-09-02
Never send an empty content block

### LINKED ISSUE #762: `litellm.completion` errors with prompt cache metadata for empty assistant message
This came up in a long-running session this afternoon. Will figure out and add repro details later.

Note: this may be specific to the first-party Anthropic API.

eb8680:
Resolves #762 

This PR prevents the `handlers.llm` harness from constructing or accepting messages with empty content blocks, which provider APIs (especially Anthropic) don't like and don't respond to uniformly. This also prevents the failure in #762 from happening.

### DIFF
--- a/effectful/handlers/llm/harness/durability/transaction.py
+++ b/effectful/handlers/llm/harness/durability/transaction.py
@@ -43,9 +43,12 @@ class HistoryBuilder(ObjectInterpretation):
         they sit here: every message the harness records passes through this
         method, including the ones a failed attempt records on its way out, and
         those are the ones that get a history into a shape no provider will
-        accept. The third holds by construction -- `to_content_blocks` and
-        `_render_prompt_section` are the only sources of blocks and neither
-        builds an empty one -- so a violation is a bug in a producer.
+        accept.
+
+        The third check is about content. It holds already: the only sources of
+        content blocks are `to_content_blocks` and `_render_prompt_section`, and
+        neither builds an empty one. It is here to catch a producer that stops
+        holding to that.
         """
         history = cls.get_history()
         assert cls._carries_no_empty_block(message), (
@@ -62,7 +65,11 @@ class HistoryBuilder(ObjectInterpretation):
 
     @staticmethod
     def _carries_no_empty_block(message: Message) -> bool:
-        """Whether `message` is free of the empty text blocks Anthropic rejects."""
+        """Whether `message` is free of the empty text blocks Anthropic rejects.
+
+        String content is not checked. An empty string is a message with nothing
+        in it, which is a different problem from a block with nothing in it.
+        """
         content = message.get("content")
         return not isinstance(content, list) or not any(
             _is_empty_text_block(block) for block in content
--- a/effectful/handlers/llm/harness/durability/transaction.py
+++ b/effectful/handlers/llm/harness/durability/transaction.py
@@ -13,7 +13,10 @@ from effectful.handlers.llm.harness.hooks import (
     call_tool,
     call_user,
 )
-from effectful.handlers.llm.harness.serialization import ToolCallID
+from effectful.handlers.llm.harness.serialization import (
+    ToolCallID,
+    _is_empty_text_block,
+)
 from effectful.ops.semantics import fwd, handler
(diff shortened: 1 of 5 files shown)


effectful/handlers/llm/harness/durability/transaction.py
```diff
@@ -36,12 +39,18 @@ def get_history(cls) -> collections.abc.MutableSequence[Message]:
     def append_message(cls, message: Message) -> None:
         """Append `message` to the ambient history, if it is legal where it lands.
 
-        Both checks are about position rather than content, which is why they sit
-        here: every message the harness records passes through this method,
-        including the ones a failed attempt records on its way out, and those are
-        the ones that get a history into a shape no provider will accept.
+        Two of the checks are about position rather than content, which is why
```
jfeser: The claude comments are a rough read. I don't know if there's an easy way to get claude to use normal english, but I've had better luck with getting chatgpt to produce reasonable text.

effectful/handlers/llm/harness/hooks.py
```diff
@@ -267,7 +267,10 @@ def call_assistant[T](
     result = None
     if not tool_calls:
         serialized_result = message.get("content") or message.get("reasoning_content")
-        assert isinstance(serialized_result, str)
+        assert isinstance(serialized_result, str) and serialized_result.strip(), (
```
jfeser: I don't know if we can assert that model results are nonempty. Empty results should be retried like other invalid outputs.

effectful/handlers/llm/harness/serialization.py
jfeser, line 179: Per chatgpt, if we change the semantics of `to_content_blocks` to return an empty list when `encoded_obj == ''`, we lose output in some cases where formatting would have produced non-empty content (e.g. `{value:r}`).

effectful/handlers/llm/harness/serialization.py
```diff
@@ -113,6 +116,18 @@ def walk(v: typing.Any) -> None:
     return blocks
 
 
+def _is_empty_text_block(block: typing.Any) -> bool:
```
jfeser: This test is more restrictive than the test for empty blocks in `to_content_blocks`, so `to_content_blocks` will still return blocks that fail this.

jfeser (changes requested):
A few chatgpt comments:

effectful/handlers/llm/harness/hooks.py
eb8680: They are retried within `completion` when `completion` is given `num_retries > 0` as a kwarg, which the harness does by default via `LiteLLMConfigurer`. If an empty result survives to this point it's almost certainly a persistent error with the API that doesn't have a client-side workaround and should thus fail loudly.

jfeser: `TenacityRetryer` only catches `ToolDecodingError` and `ResultDecodingError`. I see that we convert `Exception` to these in a few places, but I don't think this assert is inside one of these converting `try-except`s.

effectful/handlers/llm/harness/serialization.py
eb8680: OK that should be fixed

effectful/handlers/llm/harness/durability/transaction.py
eb8680: I've de-slopped, I can't get it to stop doing that no matter how much I berate it...

effectful/handlers/llm/harness/hooks.py
eb8680: I'm not sure about putting this assertion under the retry loop because I think the only way it can arise apart from some opaque server-side error is from context truncation, but I've put it there anyway and we can see if it causes any problems down the line.

effectful/handlers/llm/harness/serialization.py
eb8680: Fixed, the same predicate is used everywhere this check is applied

effectful/handlers/llm/harness/durability/transaction.py
jfeser: Yeah, I think it's a fundamental issue with Claude.

PR #770, opened 2026-09-02
Fall back rather than fail on non-`Encodable` response types

### LINKED ISSUE #763: `call_assistant` errors on `_NoEncoding` response types
The following code errors:
```py
import litellm
import pydantic
from effectful.ops.types import Interpretation
from effectful.handlers.llm import Encodable

litellm.completion(model="gpt-4o-mini", messages=[{"role": "user", "content": "hi"}], response_format=pydantic.create_model("Box", __config__={"extra": "forbid"}, value=Encodable[Interpretation]))
```
Specifically, it raises:
```
litellm.exceptions.APIConnectionError: litellm.APIConnectionError: APIConnectionError: OpenAIException - `effectful.ops.types.Interpretation` has no `Encodable` encoding, so a value of it can be sent to the model but not decoded from the model's output. Register one with `TypeToPydanticType.register` if the model needs to produce these.
```
As a result, `call_assistant` fails for any `response_type` that emits `_NoEncoding`, even when tool calls are required.

eb8680:
Resovles #763 

This PR prevents the failure in #763 by falling back to a trivial validation schema. `call_assistant` will still fail for such return types unless `tool_choice="required"`, but when that is set correctly `Skill` calls with arbitrary return types should work as expected.

### DIFF
--- a/effectful/handlers/llm/harness/serialization.py
+++ b/effectful/handlers/llm/harness/serialization.py
@@ -434,23 +434,6 @@ def _pydantic_type_str[T](ty: type[T]) -> type[T]:
     return ty
 
 
-@dataclasses.dataclass(frozen=True)
-class _NoEncoding:
-    """Refuses a *validation* JSON schema for a type that has no encoding."""
-
-    ty: typing_extensions.TypeForm
-
-    def __get_pydantic_json_schema__(self, schema, handler):
-        if handler.mode == "validation":
-            raise pydantic.errors.PydanticInvalidForJsonSchema(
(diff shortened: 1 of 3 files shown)


jfeser (changes requested):
Per chatgpt:

After this change, tools that take unecodable parameters will appear to be callable using json. Previously, the json encoding mode checked whether parameters are serializable and disabled the tool.

eb8680:
OK, it should be fixed

eb8680:
Let's land #771 before this, then I'll update it to get rid of `WithJsonSchema` as in #771

PR #771, opened 2026-09-02
Generate and forward recursive schemas for recursive types

### LINKED ISSUE #761: Schema generation fails for recursive type aliases
The following code
```py
import pydantic
from effectful.handlers.llm.harness.serialization import _inline_refs

type A[T] = T | list[A[T]]
_inline_refs(pydantic.TypeAdapter(A).json_schema())
```
fails with a `RecursionError`:
```
...
  File "/Users/eli/development/effectful/effectful/handlers/llm/harness/serialization.py", line 305, in _resolve
    return [_resolve(item) for item in obj]
            ~~~~~~~~^^^^^^
RecursionError: maximum recursion depth exceeded
```
This is a bug in the `_inline_refs` helper.

eb8680:
Resolves #761 

This PR removes the wonky inlining logic for schemas with references (which are ubiquitous in Pydantic output) in favor of simply renaming relative references to absolute ones at the appropriate points. This has the nice side effect of correctly handling recursive schemas like the failure case in #763.

### DIFF
--- a/effectful/handlers/llm/harness/serialization.py
+++ b/effectful/handlers/llm/harness/serialization.py
@@ -10,7 +10,6 @@ import collections.abc
 import contextvars
 import dataclasses
 import functools
-import hashlib
 import inspect
 import io
 import json
@@ -29,7 +28,6 @@ from litellm import (
     ChatCompletionToolParam,
     OpenAIMessageContentListBlock,
 )
-from openai.lib._pydantic import _ensure_strict_json_schema
 from openai.types.chat import (
     ChatCompletionMessageToolCall as OpenAIChatCompletionMessageToolCall,
 )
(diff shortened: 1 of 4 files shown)


jfeser (changes requested):
Per chatgpt, hashing by the schema definition can cause us to conflate distinct definitions:

In `module_a.py`:
```
class Leaf(BaseModel):
    value: int
class Parent(BaseModel):
     leaf: Leaf
```

In `module_b.py`:
```
class Leaf(BaseModel):
    value: str
class Parent(BaseModel):
     leaf: Leaf
```

In `module_c.py`:
```
class Response(BaseModel):
    int_val: module_a.Parent
    str_val: module_b.Parent
```

I'm worried that we're starting to replicate one of the more complex parts of json schema handling, which previously we had delegated to `pydantic`. Chatgpt suggests that the use of `pydantic.WithJsonSchema` is problematic because it means that `pydantic` can't do this resolution for us.

eb8680:
> I'm worried that we're starting to replicate one of the more complex parts of json schema handling, which previously we had delegated to pydantic

This PR is simpler and more correct than what's in `master` as evidenced by the failure in #761 

> Chatgpt suggests that the use of pydantic.WithJsonSchema is problematic because it means that pydantic can't do this resolution for us.

Pydantic [does not do the necessary resolution at all for user-generated schemas](https://github.com/pydantic/pydantic/issues/12145), hence the need for a workaround. Moving away from `WithJsonSchema` would be a much bigger change. I'm not even sure how that could work at all given that we really do need to generate a schema for the LLM API, not just validate and serialize.

jfeser:
Chatgpt suggests using `pydantic`'s `core_schema`. This is (I think) the intermediate representation that it uses for resolution. I haven't spent a lot of time with this problem, so I can't be personally confident about the right direction.

eb8680:
> Chatgpt suggests using pydantic's core_schema. This is (I think) the intermediate representation that it uses for resolution.

What does this mean? Slop in the form of runnable failing test cases is one thing but I can't make heads or tails of this suggestion. AFAIK the only mechanism other than `WithJsonSchema` for customizing schema generation for a class is defining a `__get_pydantic_core_schema__` method on the class (maybe this is what you're referring to?), which is not applicable in this setting where we do not necessarily own or control the types we are handed, e.g. the way we use `PIL.Image.Image` to represent images. Even when you can implement it, that mechanism does nothing for the original problem of dangling user-defined references in compositional schema generation - the open Pydantic issue referenced above uses this API in its failure case.

eb8680:
You can see the relevant history of this part of the codebase in #584 and #619 , the design in `master`/#694 is more or less the same as what was converged on there.

eb8680:
I found a different workaround that gets rid of the explicit schemas and associated schema-munging. It seems `pydantic.BeforeValidator` and `pydantic.PlainSerializer` now take arguments specifying a type from which to derive a schema, e.g. a `BaseModel` subclass. I migrated to that API and removed all `pydantic.WithJsonSchema` usage. Now #761 is fixed and the code is notably simpler.

PR #776, opened 2026-09-02
Switch to stronger model in CI builds

### LINKED ISSUE #773: LLM example tests are flaky with weak model
`gpt-4o-mini` is a bad model and often exceeds retry limits on basic examples, leading to spurious CI failures. We should switch to a different model or use a fallback in CI.

eb8680:
Resolves #773 (hopefully)

### DIFF
--- a/.github/workflows/nightly_examples.yml
+++ b/.github/workflows/nightly_examples.yml
@@ -18,7 +18,7 @@ on:
       model:
         description: "Model to run the examples against"
         required: false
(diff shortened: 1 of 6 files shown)


eb8680:
Two empty responses in the CI failures, not sure what that's about. Maybe context limits?

eb8680:
It seems the problem is `gpt-5-mini` is a reasoning model, so the token budgets are wrong.

datvo06 (approved):

